PR #32, opened 2022-06-21
Eslint setup

PDXChristian:
# Issue

We have rules for styling. These rules were often ignored because we didn't have any hard guidance. 

# Fix

I have implemented ESLint following the Google styling guide. This will throw an error when you compile. I will be adding support for this with Github Actions in the future to prevent merges without fixing linter issues.


apj640 (approved):
looks good

Issue #115, opened 2022-10-14
Fix showtimes/ticket selection on main page


arasmu2:
After clicking on the 'See Showings' button next to an event on the main page, the drop down fields for date/time/ticket selection are no longer populating correctly. Will require front/back end work to modify SQL queries and handle their results properly.

arasmu2:
I fixed the date parsing on the bac end and the site now displays the correct dates. 

The time is being correctly parsed on the back end, but is being displayed incorrectly still

arasmu2:
Definitely going to need some front end assistance, this whole ticket selection mechanism is going to need refactoring

PR #116, opened 2022-10-14
DB, ticket router update

arasmu2:
## Brief Description

Added instructions to connect PGAdmin to the docker DB

Updated the DB to mark only instances from the current season as active

Fixed the SQL query that returns tickets to active instances
-for rows returned by this query, fixed the date parsing to account for the new DB

## References Issue

Partial resolution of Issue #115

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation

### Breaking Changes

None

## Testing done

- [x] Docker was used.

**System Specifications**:
* Operating System:
* Node Version:
* NPM Version:

## Checklist:

- [x] I have used the linter and fixed any linting issues.
- [ ] I have adjusted the documentation to match the changed code.


task_sql/notes.txt
```diff
@@ -1,3 +1,20 @@
...
+  Open PGAdmin and select 'Add New Server'
+    Under the 'General' tab, give the server a name
+    Under the 'Connection' tab:
+      Set hostname/address to 'localhost'
+      Verify Port is set to 5432
+      Set username/password to the same values in your .env file (postgres default for both)
+  Click save and you will be connected
+
+
+To connect to the DB from docker command line:
```
ben16williams, line 18: Docker command line is a little ambiguous, I would suggest elaborating here more e.g. "From Docker Desktop, go to the server container and select 'container actions' -> 'open in terminal'" or something like that

arasmu2:
@ben16williams pinging you for a review/approval

server/src/api/events/event.service.ts
```diff
@@ -5,38 +5,37 @@ import {pool, response, buildResponse} from '../db';
...
+          eventid id, 
           seasonid, 
           eventname title, 
           eventdescription description, 
           active, 
           imageurl image_url, 
           count(eventinstances.eventinstanceid) as numShows
-          FROM events natural join eventinstances
-          GROUP BY events.eventid,
+        FROM events natural join eventinstances
```
ben16williams, line 16: I think we should avoid using NATURAL JOIN, it makes reading the query more difficult unless you know all the fields of the table. I think an equivalent command would be `FROM eventinstances JOIN events ON eventid` if the schema I'm looking at is current. The commands should also be capitalized for consistency

server/src/api/events/event.service.ts
```diff
@@ -201,37 +216,34 @@ export const checkIn = async (params: any): Promise<response> => {
...
+          eventinstanceid,
+          eventid,
+          eventname,
+          eventdescription,
+          imageurl,
+          eventdate, 
+          eventtime,
+          totalseats,
+          availableseats
+        FROM eventinstances NATURAL JOIN events
```
ben16williams, line 229: Same here

ben16williams (changes requested):

ben16williams:
This is totally optional but you could also modify setup.py so that the default database name is `wondertix`

server/src/api/events/event.service.ts
arasmu2: I agree, it's probably a better idea to be explicit about the joins.

arasmu2:
Can I get these changes merged for the moment? I'm updating the field names in all tables to explicitly denote a field as a foreign keys (eg: `eventid_fk`) which will require explicit joins instead of natural joins. I will also set the default DB name as `wondertix`

ben16williams:
> Can I get these changes merged for the moment? I'm updating the field names in all tables to explicitly denote a field as a foreign keys (eg: `eventid_fk`) which will require explicit joins instead of natural joins. I will also set the default DB name as `wondertix`

Sure, just make sure to put it in a ticket

ben16williams (approved):

Issue #118, opened 2022-10-15
SQL query updates


arasmu2:
Update all existing SQL to conform with the new DB design. SQL commands should be all caps. Styling examples are given below.

Queries that can fit on one line:
```
SELECT field FROM table WHERE condition = condition;
```

Queries that span multiple lines:
Simple:
```
UPDATE table 
SET field = value 
WHERE field2 = value2;
```

Complex:
```
SELECT
  field1,  
  field2,  
  ....
FROM
  table1 JOIN table2 ON table1.field = table2.field,
  ...
WHERE
  condition = condition,
  ...
GROUP BY
  field1,
  field2;
```

PR #123, opened 2022-10-20
SQL updates for new DB

arasmu2:
## Brief Description

Updated SQL queries for the following routers:
- Tickets
- Tasks
- Newsletter
- Events

Added a column to the eventinstances table that allows a default ticket type to be set for the instance

Fixed my earlier attempt to set 'wondertix' as the default database name

## References Issue

Partial resolution of issue #118

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Major change (potentially breaking, list anything that is broken below)
- [ ] Documentation

### Breaking Changes

There are multiple new SQL queries that will break requests, either by requiring more/less/different data to be provided, or will end up returning more/fewer datapoints. Will require some front end assistance to resolve.

## Testing done

- [x] Docker was used.

**System Specifications**:
* Operating System:
* Node Version:
* NPM Version:

## Checklist:

- [s] I have used the linter and fixed any linting issues.
- [x] I have adjusted the documentation to match the changed code.


arasmu2:
Renamed the following tables:
customers -> contacts
customerorders -> orders

ben16williams (commented):
Is Kendall working on customer queries and you're just renaming the table this this PR?

arasmu2:
Yes, I just renamed the tables. I told him what i was doing and he'll either do the same renaming process on his end or pull these changes before he pushes anything

server/src/api/events/event.service.ts
```diff
@@ -4,59 +4,72 @@ import {pool, response, buildResponse} from '../db';
...
-          seasonid,
-          eventname title,
-          eventdescription description,
-          active,
-          imageurl
-        FROM events
-        WHERE eventid = $1;`,
+          SELECT
+            eventid,
+            seasonid,
```
ben16williams, line 63: This query's not working, shouldn't this be `seasonid_fk` now?

server/src/api/events/event.service.ts
```diff
@@ -114,23 +127,35 @@ export const updateInstances = async (
...
+            eventdescription, 
+            active, 
+            seasonticketeligible, 
+            imageurl)
+            = ($1, $2, $3, $4, $5, $6)
+          WHERE 
+            eventid = $7
+          RETURNING *;`,
     values: [
       params.seasonid,
```
ben16williams, line 160: `seasonticketeligible` needs added as a value

server/src/api/events/event.service.ts
```diff
@@ -196,11 +228,19 @@ export const archivePlays = async (params: any): Promise<response> => {
...
+                seasonid_fk,  
+                eventname, 
+                eventdescription, 
+                active,
+                seasonticketeligible, 
+                imageurl)
+          VALUES 
+            ($1, $2, $3, true, $4, $5)
+          RETURNING *;`,
     values: [params.eventName, params.eventDesc, params.imageUrl],
```
ben16williams, line 244: seasonid_fk and seasonticketeligible need added as values

server/src/api/tickets/ticket.service.ts
```diff
@@ -42,30 +42,38 @@ const reduceToTicketState = (res: any, t: Ticket) => {
...
+export const removeTicketType = async (id: string): Promise<response> => {
   const myQuery = {
-    text: 'DELETE FROM tickettype WHERE id = $1',
+    text: `
+          UPDATE 
+            tickettype 
+          SET
+            depracated = true
+          WHERE 
+            tickettypeid = $1;`,
```
ben16williams, line 75: We should leave this as a `DELETE` since this service function is called from a `DELETE` route. That or change the route to a `PUT` instead.

ben16williams (changes requested):
Just a few small comments

arasmu2:
Thanks for catching these, I'm fixing them in my next push

server/src/api/tickets/ticket.service.ts
arasmu2: I'm going to make this route a `PUT`

arasmu2:
OK I fixed the errors you pointed out, with the exception of changing that one route to a put. 

I was working on the other routers when I realized it might be better to create a more appropriately named branch for the ticket/event/newsletter router fixes, because they're going to require some front end work to resolve the changes and having a branch dedicated to that work makes the collaboration easier. 

If these changes look good I'd like to merge to dev, then create a branch for that work

ben16williams (approved):
Sounds good!

PR #135, opened 2022-11-07
DB updates

arasmu2:
## Brief Description

Updated the database to remove a reference loop inside the seasontickets table.

Removed all references to the old customers table (renamed to contacts)

In the contacts table, changed custaddress to address (changed all references)

Updated database schema to reflect changes

Added database documentation file

## References Issue

Issue #

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] Major change (potentially breaking, list anything that is broken below)
- [x] Documentation

### Breaking Changes

None

## Testing done

- [x] Docker was used.

**System Specifications**:
* Operating System:
* Node Version:
* NPM Version:

## Checklist:

- [x] I have used the linter and fixed any linting issues.
- [x] I have adjusted the documentation to match the changed code.
- [x] Existing tests pass locally with changes.

## Additional information: 

DB updates will require a `docker compose down -v` in order for changes to take effect


server/src/api/contacts/contacts.service.ts
```diff
@@ -11,7 +11,7 @@ export const findAll = async (params: any): Promise<response> => {
             WHERE ($1::text IS NULL OR LOWER(firstname) LIKE $1)
             AND ($2::text IS NULL OR LOWER(lastname) LIKE $2)
             AND ($3::text IS NULL OR LOWER(email) LIKE $3)
-            AND ($4::text IS NULL OR LOWER(custaddress) LIKE $4)
+            AND ($4::text IS NULL OR LOWER(address) LIKE $4)
             AND ($5::text IS NULL OR LOWER(donorbadge) LIKE $5)
             AND ($6::boolean IS NULL OR seatingaccom = $6)
             AND ($7::boolean IS NULL OR vip = $7)
```
ben16williams, line 17: Do you mind fixing this query in this PR? `phone` is missing from the SQL query and `seatingaccom` is missing from the values. While we're at it, we might as well change `seatingaccom` to be a boolean in the database. I am pretty certain that changing it's type wont require any frontend changes (the frontend is already handling like other boolean values)

server/src/api/contacts/contacts.service.ts
```diff
@@ -49,7 +49,7 @@ export const findByName = async (firstname: string, lastname: string): Promise<r
 
 export const find = async (id: string): Promise<response> => {
   const myQuery = {
-    text: 'SELECT * FROM contacts WHERE customerid = $1',
+    text: 'SELECT * FROM contacts WHERE contactid = $1',
```
ben16williams, line 52: Works without the semi-colon but should add it anyways for consistency

server/src/api/contacts/contacts.service.ts
```diff
@@ -62,14 +62,14 @@ export const create = async (r: any): Promise<response> => {
...
-    values: [r.firstname, r.lastname, r.email, r.custaddress, r.phone, r.donorbadge, r.seatingaccom, r.newsletter, r.vip, r.volunteer_list],
+    values: [r.firstname, r.lastname, r.email, r.address, r.phone, r.donorbadge, r.seatingaccom, r.newsletter, r.vip, r.volunteer_list],
   };
   return await buildResponse(myQuery, 'POST');
 };
 
 export const remove = async (id: string): Promise<response> => {
   const myQuery = {
-    text: 'DELETE FROM contacts WHERE customerid = $1;',
+    text: 'DELETE FROM contacts WHERE contactid = $1;',
```
ben16williams, line 72: This delete won't work as is because of the fk policy linking the `contacts` and `orders` tables. We don't want to delete order information but there should be an option to delete contact information in case a user doesn't want there data to be stored anymore. There is also functionality in the front-end for admins that calls the related delete endpoint. So to avoid a refactor and to give users at least some kind of option to have there data removed (even if they have to contact PPH to do so), my suggestion is to set all fields to `NULL` except the `custid`. Also, so that contacts that have been nulled out do not show up in reporting, we should add logic to only (for now) to the `findAll`  service function that only returns results that have a non-null email address since contacts are required to have an email address. This could probably be implemented directly into the query somehow, but it is more complex so I'm not sure exactly.

ben16williams (changes requested):

ben16williams:
It's not a blocker but what is the reason for such a large dif in `mytix_dump.sql`?

arasmu2:
> It's not a blocker but what is the reason for such a large dif in `mytix_dump.sql`?

I realized while writing up the documentation there were a few fields that needed to be marked as `NOT NULL` in addition to changing a few foreign key relations. I opted to rebuild the DB rather than trying to modify everything in place

server/src/api/contacts/contacts.service.ts
arasmu2: `seatingaccom` is already a boolean in the DB, will add `phone` to the query in my next push

server/src/api/contacts/contacts.service.ts
arasmu2: Fixed in next push

server/src/api/contacts/contacts.service.ts
arasmu2: I assume the front end handles the requirement for an email address at the moment?

ben16williams: Yes, an email address is required when creating a contact along with some other fields

server/src/api/contacts/contacts.service.ts
```diff
@@ -11,24 +11,27 @@ export const findAll = async (params: any): Promise<response> => {
...
             AND ($2::text IS NULL OR LOWER(lastname) LIKE $2)
             AND ($3::text IS NULL OR LOWER(email) LIKE $3)
-            AND ($4::text IS NULL OR LOWER(custaddress) LIKE $4)
-            AND ($5::text IS NULL OR LOWER(donorbadge) LIKE $5)
-            AND ($6::boolean IS NULL OR seatingaccom = $6)
-            AND ($7::boolean IS NULL OR vip = $7)
-            AND ($8::boolean IS NULL OR volunteerlist = $8)
-            AND ($9::boolean IS NULL OR newsletter = $9)`,
+            AND ($4::text IS NULL OR LOWER(address) LIKE $4)
+            AND ($4::text IS NULL OR LOWER(phone) LIKE $5)
```
ben16williams, line 15: Duplicate fours

ben16williams (changes requested):

server/src/api/contacts/contacts.service.ts
ben16williams: Oops I think I meant `donorbadge` can be changed to a boolean

server/src/api/contacts/contacts.service.ts
arasmu2: fixed

server/src/api/contacts/contacts.service.ts
arasmu2: `donorbadge` is changed to a boolean now

ben16williams (approved):

Issue #42, opened 2022-07-14
When given an invalid ID, API endpoints still return status 204


apj640:
STEPS TO RECREATE:

1. Find an ID that does not exist in one of the tables, such as Donations. (Alternatively, delete a row from the table and use the deleted ID).
2. Make a request to an endpoint that takes an ID and uses the table from previous step, passing in the ID that does not exist.
3. Observe the status code the endpoint sends back

FIX:
When given an invalid ID, endpoint should return status 404, because the requested resource does not exist.

PR #136, opened 2022-11-09
200to404

KENDARRR:
## Brief Description

When invalid path request were sent to the server, the server would reply with 200/204 and an empty list. Now invalid paths return 404, as expected.

Issue #42 

## Type of change

- [ !] Bug fix
- [ ] Feature
- [ ] Documentation

### Breaking Changes

A some unit tests were briefly broken, but I believe I fixed them all. The mocked functions were replying with success and empty lists. So fixing the issue above caused the unit tests to fail. The issue was fixed by adding a 0 to the returned list, thus it is no longer empty.

## Testing done

Describe tests that were used, along with instructions for testing.

- [! ] Docker was used.

**System Specifications**:
* Operating System:
* Node Version:
* NPM Version:

## Checklist:

- [ !] My code follows the styling guidelines.
- [ !] I have added unit tests and verified that they pass.
- [ ] I have used the linter and fixed any linting issues.
- [ !] I have commented the code.
- [ ] I have adjusted the documentation to match the changed code.
- [ ] Prior to submitting the PR, I made sure the latest changes were merged with my branch.

### DIFF
--- a/server/src/api/accounts/accounts.router.ts
+++ b/server/src/api/accounts/accounts.router.ts
@@ -21,7 +21,11 @@ accountsRouter.get('/', async (req: Request, res: Response) => {
 accountsRouter.get('/search', async (req: Request, res: Response) => {
   try {
     const resp = await findByUsername(req.query.username as string);
-    const code = resp.status.success ? 200 : 404;
+    let c1 = resp.status.success ? 200 : 404;
+    if(resp.status.success === true && resp.data.length === 0){
+      c1 = 404;
+    }
+    const code = c1;
     res.status(code).send(resp);
   } catch (err: any) {
     res.status(500).send(err.message);
--- a/server/src/api/accounts/accounts.router.ts
+++ b/server/src/api/accounts/accounts.router.ts
@@ -35,7 +35,11 @@ accountsRouter.get('/search', async (req: Request, res: Response) => {
 accountsRouter.get('/:id', async (req: Request, res: Response) => {
   try {
     const resp = await find(req.params.id);
-    const code = resp.status.success ? 200 : 404;
+    let tempc = resp.status.success ? 200 : 404;
+    if(tempc === 200 && resp.data.length === 0){
+      tempc = 404;
+    }
+    const code = tempc;
     res.status(code).send(resp);
   } catch (err: any) {
     res.status(500).send(err.message);
@@ -55,7 +59,11 @@ accountsRouter.post('/', async (req: Request, res: Response) => {
 accountsRouter.delete('/:id', async (req: Request, res: Response) => {
   try {
     const resp = await remove(req.params.id);
-    const code = resp.status.success ? 204 : 404;
+    let tempc = resp.status.success ? 200 : 404;
+    if(tempc === 200 && resp.data.length === 0){
+      tempc = 404;
+    }
+    const code = tempc;
     res.status(code).send(resp);
   } catch (err: any) {
     res.status(500).send(err.message);
@@ -65,7 +73,11 @@ accountsRouter.delete('/:id', async (req: Request, res: Response) => {
 accountsRouter.put('/:id', async (req: Request, res: Response) => {
   try {
     const resp = await update(req.body, req.params.id);
-    const code = resp.status.success ? 200 : 404;
+    let tempc = resp.status.success ? 200 : 404;
+    if(tempc === 200 && resp.data.length === 0){
+      tempc = 404;
+    }
+    const code = tempc;
     res.status(code).send(resp);
   } catch (err: any) {
     res.status(500).send(err.message);
--- a/server/src/api/accounts/accounts.router.ts
+++ b/server/src/api/accounts/accounts.router.ts
@@ -35,12 +35,11 @@ accountsRouter.get('/search', async (req: Request, res: Response) => {
 accountsRouter.get('/:id', async (req: Request, res: Response) => {
   try {
     const resp = await find(req.params.id);
-    let tempc = resp.status.success ? 200 : 404;
-    if(tempc === 200 && resp.data.length === 0){
-      tempc = 404;
+    let code = resp.status.success ? 200 : 404;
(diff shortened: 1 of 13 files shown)


server/src/api/accounts/accounts.router.ts
```diff
@@ -21,7 +21,11 @@ accountsRouter.get('/', async (req: Request, res: Response) => {
 accountsRouter.get('/search', async (req: Request, res: Response) => {
   try {
     const resp = await findByUsername(req.query.username as string);
-    const code = resp.status.success ? 200 : 404;
+    let c1 = resp.status.success ? 200 : 404;
+    if(resp.status.success === true && resp.data.length === 0){
+      c1 = 404;
+    }
+    const code = c1;
     res.status(code).send(resp);
```
ben16williams, line 29: You could drop `c1` and just use `code`. The reassigning seems unnecessary

server/src/api/discounts/discounts.router.ts
```diff
@@ -86,7 +86,11 @@ discountsRouter.put('/:id', checkJwt, checkScopes, async(
 ) => {
     try {
       const oldCode = await alterDiscountCode(req.params.id);
```
ben16williams, line 88: We should refactor `oldCode` to `resp` everywhere

server/src/api/accounts/accounts.router.ts
KENDARRR: The originals returned const and wasn't sure if that needed to be maintained, but [I found articles](https://medium.com/dailyjs/use-const-and-make-your-javascript-code-better-aac4f3786ca1) saying it was at the very least recommended.

server/src/api/discounts/discounts.router.ts
KENDARRR: Fixed

server/src/api/accounts/accounts.router.ts
ben16williams: Generally yes but in this instance the scope that the variable is in is small and having multiple differently named variables for the same value makes it harder to read so I think leaving it has `let` would be fine. You could also extract a function like the article suggests which would also be good future proofing if add any more return scenarios in the future

KENDARRR: I'm really confused, did I not do that? One of the conditionals is a ternary operator/assignment, but it's otherwise the same, no?

ben16williams: I'm not sure what you mean, the ternary operator is fine. I'm talking about doing either 
```
let code = resp.status.success ? 200 : 400;
if ( resp.status.success === true && resp.data.length === 0) {
    code = 404;
}
res.status(code).send(resp);
```
or like the article suggests for extracting a function like
```
const calculateCode = () => {
    if (resp.status.success === true) {
        if (resp.data.length === 0) {
            return 404;
        }
        return 200;
    }
    return 404;
}
const code = calculateCode();
```

ben16williams: In both of these `c1` is dropped which simplifies things because there is one less variable to keep track of and there is no real security risks having `code` be mutable like in the first code block

KENDARRR: Alright I'll just get rid of const then

ben16williams (commented):
Not sure if you're done updating all the routes, if not ignore my recent comments

server/src/api/accounts/accounts.router.ts
```diff
@@ -21,7 +21,11 @@ accountsRouter.get('/', async (req: Request, res: Response) => {
 accountsRouter.get('/search', async (req: Request, res: Response) => {
   try {
     const resp = await findByUsername(req.query.username as string);
-    const code = resp.status.success ? 200 : 404;
+    let c1 = resp.status.success ? 200 : 404;
+    if(resp.status.success === true && resp.data.length === 0){
+      c1 = 404;
+    }
+    const code = c1;
```
ben16williams, line 28: Should just need to change over a couple more

server/src/api/contacts/contacts.router.ts
```diff
@@ -54,7 +58,12 @@ contactsRouter.get('/:id', async (req: Request, res: Response) => {
 contactsRouter.delete('/:id', async (req: Request, res: Response) => {
   try {
     const resp = await remove(req.params.id);
-    const code = resp.status.success ? 204 : 404;
+    let tempc = resp.status.success ? 200 : 404;
+    if(tempc === 200 && resp.data.length === 0){
+      tempc = 404;
+      resp.status.success = false;
+    }
+    const code = tempc;
```
ben16williams, line 66: here

server/src/api/discounts/discounts.router.ts
```diff
@@ -104,9 +109,14 @@ discountsRouter.delete('/:id', checkJwt, checkScopes, async(
...
-      const code = oldCode.status.success ? 200 : 404;
-      res.status(code).send(oldCode);
+      const resp = await deleteDiscountCode(req.params.id);
+      let tempc = resp.status.success ? 200 : 404;
+      if(tempc === 200 && resp.data.length === 0){
+        tempc = 404;
+        resp.status.success = false;
+      }
+      const code = tempc;
+      res.status(code).send(resp);
```
ben16williams, line 119: here

server/src/api/events/event.router.ts
```diff
@@ -396,7 +400,11 @@ eventRouter.put('/instances/:id', checkJwt, checkScopes, async (
 ) => {
   try {
     const resp = await updateInstances(req.body, req.params);
-    const code = resp.status.success ? 200 : 404;
+    let tempc = resp.status.success ? 200 : 404;
+    if(tempc === 200 && resp.data.length === 0){
+      tempc = 404;
+    }
+    const code = tempc;
```
ben16williams, line 407: here

ben16williams (approved):

PR #190, opened 2023-04-24
Swagger docs tasks

diaboloshogunate:
## Brief Description

swagger docs for tasks routes.
Also fixing a small bug in the service that is using a incorrect column name.


samkhodak (approved):
Looks good!

PR #196, opened 2023-04-25
swagger docs for refunds

diaboloshogunate:
## Brief Description

add swagger docs for refunds route.

Note this route does not work correctly


samkhodak (approved):
Looks good!

PR #203, opened 2023-05-03
Change route after adding an event

xnsc:
## Brief Description

After adding an event, clicking the pop-up confirmation's 'Proceed' button will redirect the user to the Showings page rather than the Manage Events page.

## References Issue

Issue #201

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation


samkhodak (approved):
Approved. Thank you!

PR #207, opened 2023-05-03
revert auto install of cert

diaboloshogunate:
Removed creation of cert for server.

You will need to make the cert on the host machine like before.


samkhodak (approved):
Approved. Thank you!

PR #217, opened 2023-05-16
update readme with new docker instructions

diaboloshogunate:
This is just a document update to the readme.

I am removing the instructions for local installs to simplify things. Docker is preferred to ensure everyone is on the same version of dependencies. 

Changed the headings to a simple list.

Included information about `.env` file, swagger docs and having to type `thisisunsafe`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,58 +1,41 @@
 # WonderTix
-
-For the most current stable version see the 'main' branch.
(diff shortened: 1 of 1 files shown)


samkhodak (approved):
Well done. Thanks.

PR #221, opened 2023-05-17
manageevent related code removed

cmdrpug:
Removed both files related to the manageevent page and its functionality. Imports and link to those files/pages were deleted. The nav call after saving an event now goes to the showings page.


samkhodak (approved):
Sponsor approved switching to showings only. Thank you!

PR #256, opened 2023-06-02
Ticket type buttons

phamt007:
## Brief Description

Fixed Ticket Type Button to display a certain types of tickets based on the available seats of the showing and the ticket restriction table, if else it uses the default ticket type of the showing and available seats. If there are no available seats, then there should not be a ticket allowed to be purchased.


samkhodak (approved):
Approved. Thank you!

PR #292, opened 2023-07-23
Jest fix for vscode and .gitignore setup

jgreever:
_**Many changes are from running `npm lint:fix`**_

_changes that were in the prior PR have been moved into this one (.gitignore)_

Added vscode WonderTix.code-workspace file. If you open this file in
vscode, it will open the project as a workspace, keeping the root,
client, server, db, and deploy stuff in their own easy-to-use
workspaces.

Corrected issue with Jest not running in vscode. This change does
require extra configuration to get your JWT_TOKEN and add to the .env
file, then modify the ./server/src/testSetup.ts to return your JWT_TOKEN
as token, thus doing a very simplified mock of Auth0 and jwt token
authentication without needing to actually hit the Auth0 API. Currently,
it is still setup the way it has been. The required change to do the
above required manually editing. More discussion will need to happen
with Ben W. before making any permanent changes to the current jest
testing method.

Adjusted .gitignore to be the only one (root folder) and added content
from the server/.gitignore and client/.gitignore into it.

Ran npm lint:fix

Renamed a few variables to match camelCase.

Updated server/package.json dependencies and verified no regression
errors and no deprecated functions were introduced.

Removed package-lock.json from root folder since it is not needed.


WonderTix.code-workspace
ben16williams, line 1: There is usually a wide variety of text editors/ide's that people use, but I'm not opposed to adding this. I tried it out and it looks pretty nice! There should be a comment about what it is/how to use it with VSCode in the README or inline (if comments are allowed in this file idk) otherwise it's kind of just another random file to anybody that isn't using the one text editor it's meant for

server/.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
ben16williams, line 1: Is this file needed?

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
+    "@types/node": "^20.4.4"
+  }
+}
```
ben16williams, line 5: Is this file necessary? I was able to build and run without it. Not opposed just wondering what it's for in the root

ben16williams (dismissed):
Thanks for doing this! I was able to get all the tests to pass and build/run locally, but since we don't have a dev environment to push this to and see if it breaks, it would be nice if someone else pulled it down and tried to build too

package.json
jgreever: I have deleted this file, it was a leftover artifact from testing locally, thank you for catching it!

server/.vscode/settings.json
jgreever: It is not and has been deleted! Thank you!

WonderTix.code-workspace
jgreever: Updated the README.md (and did some linting) explaining how to open the workspace.

xiivi (approved):
Looks good

jgreever:
Ok, I fixed the README.md file merge issue and push the commit back up. If you can take one more peek at it please and thank you!.

xiivi (approved):
Looks good 👌

ben16williams (approved):

PR #296, opened 2023-07-28
[TIC-12] Solution/create documentation for live updates when editing front-end

benfaustin95:
### Description
Added solution and documentation for live updates when editing front-end or backend. In addition I added a health-check to the docker-compose.yaml so the server container will not try to seed if the database fails to build, as well as updated the .env so that the environment variable names matched their counter parts being loaded in docker-compose.yaml. 

### Risks
Potential risks include potential docker configuration conflicts via changes that have not yet been merged into ticketing and/or main. I would recommend cloning TIC-12 and testing the changes before merge on a non Mac/M1 machine. 

### Validation
I validated changes by manually testing site features after changes were made.

### Issue
[*Link to corresponding issue*](https://pdx-hkaus.atlassian.net/browse/TIC-12)

### Operating System
CPU: M1 (ARM CPU)
OS: macOS 13.1


henrykaus (commented):
I will be taking more of a look into this, this week since I have been unable to come up with a solution for live updates for Ethan, Jonathan and I.

PR #298, opened 2023-07-28
[DEP-22] mkcert fixed #297

### LINKED ISSUE #297: [DEP-22] fix mkcert
Mkcert should have eliminated the Chrome Certificate Error which required a `thisisunsafe` input or the disabling of the security feature. 

Investigate and fix this issue.

jtng3:
### Description
Mkcert is fixed so that the certificate warning on Chrome will no longer appear. This is done by pointing SSL_CRT_FILE to /usr/app/localhost.pem and SSL_KEY_FILE to /usr/app/localhost-key.pem in the client package.json file.
This requires the user to copy the certificate keys from the server to the client folder for the docker builds to work.
Readme updated to reflect the need to copy the local certificate public and private keys into the client folder

### Risks
This requires the user to copy the certificate keys from the server to the client folder for the docker builds to work.

### Validation
Certificate Warning no longer appears. 

### Issue
#297 

### Operating System
Mac M Series

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,12 +28,25 @@ Future features include managing/creating email campaigns and ticket exchanges.
 ## Setup
 
 1. Clone the repository.
-2. Create a `.env` file and copy over the contents from the `.env.dist` (.env example) file (or copy `.env.dist` to `.env` and edit)
-   1. Set the values for `AUTH0_CLIENT_ID` and `AUTH0_CLIENT_SECRET`. *you can get these values from the team lead*
-   2. Set the value for `PRIVATE_STRIPE_KEY` and `PUBLIC_STRIPE_KEY`. *you can get this value from the team lead*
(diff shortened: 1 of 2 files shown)


henrykaus (approved):
This appears to fix the issue with Chrome! It still doesn't work for Firefox, but as far as development goes, that may be ok. Just something to consider when Deployment gets around to doing actual certificates in the future.

jgreever:
the client/.gitignore and server/.gitignore files are no longer needed as there is a main .gitignore in the root folder that already should be blocking localhost*.pem, cer, and key files. You are safe to ignore those two files when fixing the conflicts prior to merging in main.

henrykaus (approved):
Good!

xiivi:
for any folks who are unable to access the client's web pages after merging this PR to your local main, @raadelmsellati discovered that you're required to navigate into your client directory and execute the following command:

`mkcert localhost`

PR #299, opened 2023-07-28
Fixed DB_USER to DB_PORT

jgreever:
### Description
Fixed DB_USER to DB_PORT in the variable calling the wait-for-it.sh script since the docker logs were saying 'nc: invalid port postgres' and now it works fine since it's not being passed the ip:username, it's now the ip:port as it should be.

### Risks
None

### Validation
Passed on Docker, error not showing in the logs any more

### Issue
NA

### Operating System
macOS (intel) 13.0

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -12,4 +12,4 @@ WORKDIR /usr/app
 COPY ./package*.json .
 RUN npm install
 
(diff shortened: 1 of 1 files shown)


xiivi (approved):
Looks good 👍

Issue #238, opened 2023-05-27
Ticketing - Mobile nav x button to close nav should show as a link, not just an image.


samkhodak:
This mobile nav can be found in ticketing and admin dashboard when the user opens the hamburger menu to show the nav. 
Currently, the x 'button' doesn't show a different pointer to indicate that this X is selectable. This is probably an accessibility issue.

PR #311, opened 2023-08-01
[TIC-37]- Fixed "X" in mobile view.  Pointer now changes to hand pointer when hovering over X

txiong002:
### Description
Currently, the X 'button' doesn't show a different pointer to indicate that this X is selectable.  Fixed issue by replacing div tag with button tag in udash_navbar.tsx and Navigation.tsx files for moblie menu option id=closeSideBar.

### Risks
No risk

### Validation
In mobile view when hovering over X, the pointer now changes to finger pointer icon like the rest of icons in side bar.  No errors in index window when clicking on X. 
image

### Issue
https://github.com/WonderTix/WonderTix/issues/238

### Operating System
Windows


henrykaus (changes requested):
Everything looks good here! Good catch on the other instance of the same mobile nav bar.

I have one more spot where this same issue persists and it should be another simple `div` -> `button` switch. In the main navbar on `localhost:3000` there is a hamburger and x button. The change should be able to be made on line 104 in `Navbar.tsx`.

image

henrykaus (approved):

PR #315, opened 2023-08-04
[TIC-35] Create and add a default image to events that don't select an image

### LINKED ISSUE #247: Ticketing - Image URLs should be checked for correctness or have a default image if the URLs are broken.
Currently, event images just give a broken blank icon when the URLs are not correct. 
They should be checked for correctness when adding an event, to avoid broken images if possible.

benfaustin95:
### Description
1. Added Default Event Image to assets
2. Updated all event image uses to include fall back to default image in event of load failure
3. Updated add event form to allow user to select default image when creating event

### Risks
Conflicts with modifications being made by other developers.

### Validation
Verified changes by: 
1. Added new event using default image, checked image loaded correctly.
2. Added new event with broken image source, checked image loaded correctly. 

### Issue
[*Link to corresponding Ticket*
](https://pdx-hkaus.atlassian.net/browse/TIC-35)

closes #247 

### Operating System
M1 Macbook

### Before 
image
image

### After
image
image
image
image
image


client/src/components/Ticketing/ticketingmanager/Events/EventForm.tsx
```diff
@@ -169,48 +171,39 @@ const EventForm = ({onSubmit, tickettypes, initialValues}: EventFormProps) => {
...
-    if (showings.length === 0) {
-      setErr('Please enter at least one showing.');
+    const name = data.eventname?.trim() === '';
+    const desc = data.eventdescription?.trim() === '';
+    const count = (name?1:0)+(desc?1:0);
+    if (count) {
+      const len = count>1;
+      setErr( `The ${name?'Event Name':''}${len?', and ':''}
+      ${desc?'Event Description':''}
+      ${len?' fields are missing':' field is missing'}`);
```
henrykaus, line 181: Could you make the variable names used here a little more specific, something like `hasNoName`, `hasNoDesc`, and `missingFieldsCount`.

henrykaus: Thanks for simplifying this though, it was a big mess before

client/src/components/Ticketing/ticketingmanager/Events/InputField.tsx
```diff
@@ -9,39 +9,58 @@ interface InputFieldProps
...
+        <label
+          htmlFor={id}
+          className='text-sm text-zinc-600 '
+        >
+          {headerText}
+        </label>
+        <input
+          className='w-full p-2 rounded-lg border border-zinc-300 mb-4'
+          id={id}
+          type={type}
```
henrykaus, line 31: Will the fact that the input type changed from a set `'input'` to being passed in as a prop affect other areas where `InputField` is used?

client/src/components/Ticketing/ticketingmanager/Events/showingInputContainer.tsx
```diff
@@ -151,10 +151,16 @@ const ShowingInputContainer = ({
       <div key={index} className='shadow-xl p-5 rounded-xl mb-9 bg-violet-700'>
         <label className='font-semibold text-white mb-7 mt-7  '>Show # {index + 1}</label>
         <div className='flex flex-col gap-5 mt-5 md:pr-20'>
-          <h3 className='font-semibold text-white'>Total Tickets For Showing</h3>
+          <label
```
henrykaus, line 154: Thanks for switching over from `h3`s to `label`s! It's a big usability and accessibility help

client/src/components/Ticketing/ticketingmanager/showings/InstancesPage.tsx
```diff
@@ -46,13 +47,20 @@ const InstancesPage = () => {
         <ul className='md:grid md:grid-cols-2 md:gap-8 sm:grid sm:grid-cols-1 sm:gap-4 mt-9'>
           {allEvents.map((eventss) => (
```
henrykaus, line 48: I know this wasn't you, but do you think you could switch over the use of `eventss` to `event` if it doesn't conflict with anything.

client/src/components/Ticketing/ticketingmanager/showings/showing/showing.tsx
```diff
@@ -79,7 +80,7 @@ const Showing = () => {
   if (eventData === undefined) return <p>Whoops! Event not found</p>;
   const {title, description, imageurl, tickets} = eventData;
   return (
-    <div className='w-full h-screen overflow-x-hidden absolute bg-cover' style={{backgroundImage: `url(${imageurl})`}}>
+    <div className='w-full h-screen overflow-x-hidden absolute bg-cover' style={{backgroundImage: `url(${getImageDefault(imageurl)}),url(${getImageDefault()})`}}>
```
henrykaus, line 83: Just a curious question, but why use `getImageDefault()` twice?

client/src/utils/imageURLValidation.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+export const EventImage = (props:EventImageProps) => {
+  const {src, className, title} = props;
+  return (
+    <>
+      <img
+        className={className}
+        src={getImageDefault(src)}
+        alt={`${title} Playbill`}
+        onError={imageOnError}/>
```
henrykaus, line 28: I didn't know this prop existed, this is great!

client/src/components/Ticketing/ticketingmanager/Events/EventForm.tsx
```diff
@@ -243,21 +236,47 @@ const EventForm = ({onSubmit, tickettypes, initialValues}: EventFormProps) => {
...
+                      name={'defaultImageUrl'}
+                      id={'defaultImageUrl'}
+                      type='checkbox'
+                      value={'default'}
+                      checked={disabledUrl}
+                      onChange={ () => {
+                        setImageURL((!disabledUrl?'Default Event Image':''));
+                        setDisabledUrl(!disabledUrl);
+                      }}
+                      className={'appearance-none w-8 h-8 p-2 m-auto checked:bg-blue-500 ' +
```
henrykaus, line 275: I think it would still be a good idea to have the checkmark exist, rather than only a "blue" filled color, so I don't think `appearance-none` would be ideal. I think this change would require adding a `rounded-...` class though.

henrykaus (changes requested):

client/src/components/Ticketing/ticketingmanager/Events/EventForm.tsx
henrykaus: For consistency sake, it might be best just to use a normal checkbox like is used in other parts of the application
image

client/src/components/Ticketing/ticketingmanager/Events/InputField.tsx
benfaustin95: It should be ok, 'input' is not a defined type so the inclusion wasn't doing anything, the tag was just reverting to type text. The prop interface has type as an optional value so if not supplied the tag will just revert to type text again. Also the component is only used in the Event form and actionType is never set to onSubmit, so the entire block could be removed.

client/src/utils/imageURLValidation.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import defaultImage from '../assets/DefaultEventImage.png';
+
+export const getImageDefault = (url?:string) => {
+  if (url == undefined || url === 'Default Event Image') {
```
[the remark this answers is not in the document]
benfaustin95: Updated

client/src/utils/imageURLValidation.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface EventImageProps {
+  className:string;
+  src:string;
+  title:string;
+}
+export const EventImage = (props:EventImageProps) => {
+  const {src, className, title} = props;
+  return (
+    <>
```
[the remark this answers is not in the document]
benfaustin95: Updated

client/src/components/Ticketing/ticketingmanager/Events/EventForm.tsx
benfaustin95: Updated

client/src/components/Ticketing/ticketingmanager/showings/InstancesPage.tsx
benfaustin95: Updated

client/src/components/Ticketing/ticketingmanager/showings/showing/showing.tsx
benfaustin95: the second call returns the default image as a fall back incase the first url is broken.

client/src/components/Ticketing/ticketingmanager/Events/EventForm.tsx
benfaustin95: Updated

client/src/components/Ticketing/ticketingmanager/Events/InputField.tsx
henrykaus: Ok, sounds good. I think we should still probably keep the `onSubmit` for the time being since it doesn't necessarily hurt.

henrykaus (approved):
👍🏻 👍🏻 👍🏻 👍🏻 👍🏻

PR #316, opened 2023-08-04
[DON-67] Donation Search/Delete Routes

jgreever:
### Description
Donation endpoints for searching and deleting a donation.
Searching can be done by name, id's, donation amount, donation date, etc...

### Risks
I still need to validate that we can perform offline tests using JWT tokens and bypass Auth0 so we can verify that the routes will be secure.

### Validation
Tested them using Swagger and Postman

### Issue
https://dev.azure.com/nguyenjohnny1017/DonorManagement/_sprints/taskboard/DonorManagement%20Team/DonorManagement/Sprint%202?workitem=67

### Operating System
macOS Sonoma 14.0, Intel


jgreever:
Added stripe webhook for the donation endpoint and tested locally with successful data coming back.

Guide from Stripe being used: https://stripe.com/docs/webhooks/quickstart

xiivi (approved):
Lgtm

PR #318, opened 2023-08-06
[DON-79]

jgreever:
### Description
Removed jest and all related items from the `./server/` folder.

Deleted old `/api/1/` *.test.ts files since we are migrating to
Playwright and the current server jest tests are very difficult to mock
(JWT authorization and Auth0) without hammering the API. Playwright
tests will need to be written for the new `/api/2/` endpoints and since
we are migrating to Prisma, no further development should be happening
on the `/api/1/` endpoints, so I feel it is safe for the time being to
go without tests for the server.

The next steps will be redoing the current OpenAPI setup for the
`./server/src/server.ts` file and allow creating and using an API key
that a user can request via username/password or by sending a
pregenerated API key in the header.

Currently, this setup will (hopefully) prevent any hammering of the
Auth0 api so we don't go over and don't have to pay for any extra API
calls/month.

### Risks
Flaws in new API endpoints that are written until we get Playwright and an auth setup for testing (next step)

### Validation
Verified on local system in both docker and bare metal

### Issue
[DON-79](https://dev.azure.com/nguyenjohnny1017/DonorManagement/_sprints/taskboard/DonorManagement%20Team/DonorManagement/Sprint%202?workitem=79)

### Operating System
macOS Sonoma (14.0), Intel


xiivi (approved):
NICE

PR #320, opened 2023-08-06
Replace hard-coded values with environment variables

ben16williams:
### Description
Replaces all hard-coded values that should be environment variables and added an env var to the server for the base url. All the hard-coded values replaced were for the auth0 integration. 

### Risks
New env variables should be getting configured properly after composing up

### Validation
Tested with multiple different Auth0 applications, passes unit tests

### Issue
[https://github.com/orgs/WonderTix/projects/3/views/1?pane=issue&itemId=35176383](https://github.com/orgs/WonderTix/projects/3/views/1?pane=issue&itemId=35176383)

### Operating System
win10


jtng3 (approved):
LGTM

PR #327, opened 2023-08-08
Don donor dashboard branch

xiivi:
Description:
Added missing dependency.
> import React from 'react';


jgreever (approved):
verified it locally and it all looks good. Should we add a link in the menus to access this stub or add it when that page gets functionality?

PR #348, opened 2023-08-15
Remove flowbite dependencies

henrykaus:
### Description
Removed conflicting dependencies which required `npm i --force` in the `client` directory.

### Risks
If flowbite happens to have been used (although I think I checked everywhere), then the pages it is on will likely not work. But I don't believe it is the case. If it does happen, then this PR is easy to revert.

### Validation
Checked for any uses of `flowbite` or `flowbite-react`.

### Issue
N/A

### Operating System
Windows 11

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -4,4 +4,4 @@ WORKDIR /usr/app
 
 COPY ./package*.json .
 
-RUN npm install --force
+RUN npm install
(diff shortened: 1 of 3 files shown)


jgreever (approved):
This looks great for me. I had no issues downloading and running it fresh. Good catch on the double `cross-env` being used in the client's package.json!

PR #351, opened 2023-08-17
[TIC-56] Refactor Upper Admin Nav and other UI fixes

henrykaus:
### Description
Many small UI fixes -- such as removing random comma, making sure text does or doesn't wrap in the appropriate cases, removing unnecessary scrollbars, added margin to separate elements, and refactored the upper admin/ticketing nav to be simpler, use less nodes, and use more semantic elements.

Admin Before/After:
image
image

Nav Bar Before/After:
image
image

Contacts Before/After:
image
image

image
image

Main Page Before/After:
image
image

### Risks
Shouldn't be any.

### Validation
Much side-by-side testing to ensure that all proper functionality was not tampered with (especially with the upper nav bar).

### Issue
TIC-56

### Operating System
Windows 11


txiong002 (approved):
Looks good to me.   No issues with webpage, and all new changes work.

Issue #239, opened 2023-05-27
Ticketing - Doorlist page does not work. May need routes?


samkhodak:
The doorlist page has errors in implementation.

- List of events doesn't work correctly (improper usage of `events/list` response) 
- Doorlist doesn't return anything (improper use of `/doorlist` request. This route may not be yet established?)

PR #353, opened 2023-08-18
[TIC-45] Doorlist overhaul

0Tendo:
### Description
Overhauled doorlist page
   - Separated event & date/time dropdown
   - Fixed routes and API calls
   - Added export table ability
   - Added ability to view inactive events
   - Formatted Accommodations column to support strings
   - Added event API call for all events (instead of just active events)
   - Fixed doorlist API call

The check in guest function is still not fully functional. need to link 'Arrived' cell to the check in guest API

### Risks
Minor update to doorlist API call could affect other pages. I dont think its used anywhere else right now.

### Validation
Manual testing and console logging
image

### Issue
[*Link to corresponding issue*](https://github.com/WonderTix/WonderTix/issues/239)

### Operating System
Windows


client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -63,12 +65,14 @@ const renderCheckin = ((params: GridCellParams) =>
  * columns uses name, vip, donorbadge, accomodations, num_tickets, arrived and renders
  */
 const columns = [
-  {field: 'name', headerName: 'Name', width: 200},
+  {field: 'firstname', headerName: 'First Name', width: 120},
+  {field: 'lastname', headerName: 'Last Name', width: 120},
+  {field: 'num_tickets', headerName: 'Tickets', width: 100},
+  // TODO: Fix link from renderCheckin to checkInGuest
```
henrykaus, line 71: What does this TODO mean?

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -82,86 +86,177 @@ const DoorList = () => {
...
+  const handleEventChange = (event) => {
+    const eventId = parseInt(event.target.value);
+    setSelectedEventId(eventId);
+
+    const matchingEvents = eventListFull.filter((e) => e.eventid === eventId);
+    setAvailableTimesEvents(matchingEvents);
+  };
+
+  const handleTimeChange = (event) => {
+    const eventIId = parseInt(event.target.value);
```
henrykaus, line 131: Maybe just `eventId`?

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -63,12 +65,14 @@ const renderCheckin = ((params: GridCellParams) =>
  * columns uses name, vip, donorbadge, accomodations, num_tickets, arrived and renders
```
henrykaus, line 65: Could you update this comment

henrykaus (changes requested):
Just a bunch of minor changes in addition to what we already talked about with the strange date behavior. Otherwise, this looks good!

txiong002:
Could we display the seating accommodation instead of a T/F?  If so, may need for me to merge my branch TIC-42 first.
image

Besides this everything else looks good to me.

henrykaus:
> Could we display the seating accommodation instead of a T/F? If so, may need for me to merge my branch TIC-42 first. image
> 
> Besides this everything else looks good to me.

This might now be fixed since we changed the seed data for contacts to actual accommodations?

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -82,86 +86,189 @@ const DoorList = () => {
...
+          accommodations: row[5],
+        };
+      }).filter(Boolean);
+
+      setDoorList(doorListData);
+      const rowString = jsonData[0].row.slice(1, -1);
+      const rowParts = rowString.split(',');
+      const eventNameFromData = rowParts[7].replace(/"/g, '');
+      const eventDate = rowParts[9];
+      const eventDateObject = new Date(eventDate.replace(/(\d{4})(\d{2})(\d{2})/, '$1-$2-$3'));
```
henrykaus, line 197: This whole date thing piqued my curiosity so I looked into it a little bit. It looks like `Date` objects are very finicky. But, it appears that if you use `/`s instead of `-`s, then we get the functionality we're looking for w/o the need to add 1 to the date. So, `$1/$2/$3`. https://stackoverflow.com/questions/7556591/is-the-javascript-date-object-always-one-day-off

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -82,86 +86,189 @@ const DoorList = () => {
...
-          {date && time ? `${date}, ${time}` : `${date}${time}`}
+            ))}
+          </select>
         </div>
+        <div className="mb-3">
+          <label htmlFor="time-select" className='text-sm text-zinc-500 ml-1 mb-2 block'>Choose Time</label>
+          <select id="time-select" className="select w-full max-w-xs bg-white border border-zinc-300 rounded-lg p-3 text-zinc-600 mb-7" onChange={handleTimeChange} disabled={!selectedEventId}>
+            <option className="px-6 py-3">Select Time</option>
+            {availableTimesEvents.map((event) => {
+              const eventDateObject = new Date(event.eventdate.toString().replace(/(\d{4})(\d{2})(\d{2})/, '$1-$2-$3'));
```
henrykaus, line 244: Same as comment above

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
henrykaus: Now that I think about it though. Do we even need to put it in a `Date` object. Could we just skip from the `eventDate` to the `formattedEventDate`?

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -82,86 +86,189 @@ const DoorList = () => {
...
+  const [eventListFull, setEventListFull] = useState([]);
+  const [eventListActive, setEventListActive] = useState([]);
+  const [ticketsSold, setTicketsSold] = useState(true);
+  const [showInactiveEvents, setShowInactiveEvents] = useState(false);
 
   useEffect(() => {
-    getEvents();
+    const fetchEvents = async () => {
+      try {
+        const response = await fetch(process.env.REACT_APP_API_1_URL + '/events/list/allevents');
```
HaliTran, line 100: This looks good! I would recommend adding a check right after this fetch call to check if response.ok is true, then proceed, else throw an error. The reason is because if the response HTTP status code is 400 or 500 level which indicates error codes, it technically is a "success" and the try block will continue executing even though we don't want too.

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -82,86 +86,189 @@ const DoorList = () => {
...
       });
-      const jsonData = await response.json();
-      console.log(jsonData.data);
-
-      // doorlistData.data {id: custid, name, vip, donor: donorbadge, accomodations: seatingaccom, num_tickets, checkedin, ticketno }
-      setDoorList(jsonData.data);
-      setEventName(jsonData.data[0].eventname);
-      setDate(dayMonthDate(jsonData.data[0].eventdate));
-      setTime(militaryToCivilian(jsonData.data[0].eventtime));
+      const jsonRes = await response.json();
```
HaliTran, line 170: Same as above regarding adding a check to see if response.ok is true and changing variable name to be more specific.

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -82,86 +86,189 @@ const DoorList = () => {
...
+  const [ticketsSold, setTicketsSold] = useState(true);
+  const [showInactiveEvents, setShowInactiveEvents] = useState(false);
 
   useEffect(() => {
-    getEvents();
+    const fetchEvents = async () => {
+      try {
+        const response = await fetch(process.env.REACT_APP_API_1_URL + '/events/list/allevents');
+        const jsonRes = await response.json();
+        const jsonData = jsonRes.data as any[];
```
HaliTran, line 102: Can you change the variables names to be more specific to what the API data returns? Something like `allEventRes` and `allEventResJson` etc. For readability in the future.

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
```diff
@@ -82,86 +86,189 @@ const DoorList = () => {
...
+          setTicketsSold(false);
+          return null; // Exit early if no tickets sold
+        }
+        setTicketsSold(true);
+        return {
+          id: index,
+          firstname: row[1],
+          lastname: row[2],
+          num_tickets: row[12],
+          // TODO: hook arrived with redeemed vaue
```
HaliTran, line 184: Do you happen to know what this TODO is?

0Tendo: if we want to add functionality to check people into the database we'll have to add an API call when the 'arrived' box is checked. i removed the "TO DO" comment and will make a new issue/feature request to track it

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
0Tendo: i struggled with getting the date to format properly for a while. going straight to formattedEventDate was not an option because the object has to be a string. changing from '-' to '/' does not seem to affect the result either. 
i'm sure there's a more elegant way to do this. let me know if you want me to spend more time trying to find it.

client/src/components/Ticketing/ticketingmanager/Doorlistpage/doorlist.tsx
0Tendo: good call.

txiong002 (commented):
Should we have a  submit button when checking in customers?  
image

henrykaus:
> Should we have a submit button when checking in customers? How are we updating the list? image

At the moment, we have backlogged the ability to check users in using the doorlist

henrykaus (approved):

PR #390, opened 2023-10-08
[TIC 62] Finish Order process

benfaustin95:
### Description
Currently, only the order gets sent through the check-out process. Further tables need to be updated for each ticket sold, and tickets need to be held while payment is being processed in Stripe.

*The current frontend implementation for ticket selection does not allow for General Admission ticket purchases if there is a ticket type restriction for the event.

### Risks
- The updated ticketing schema could potentially conflict with other ongoing development efforts.
- There may be flaws in the logic for order processing.
- Some seed data has become invalid as the seed values were random and do not conform to the updated database schema.

### Validation
Manual Testing of each order processing route. 

### Issue
NA

### Operating System
Apple Silicon


server/src/controllers/eventController.service.ts
```diff
@@ -0,0 +1,309 @@
...
+
+const validateContact = (formData: checkOutForm) => {
+  return {
+    firstname: validateName(formData.firstName, 'First Name'),
+    lastname: validateName(formData.lastName, 'Last Name'),
+    email: validateWithRegex(
+        formData.email,
+        `Email: ${formData.email} is invalid`,
+        new RegExp('.*'),
+    ),
```
0Tendo, line 278: could use a more strict RegExp for email since this is the data used as the unique ID for tickets

really basic example
new RegExp('^.+@.+\\..+$')

currently you can submit the form even if its a bad email like this
image

server/src/controllers/orderController.service.ts
```diff
@@ -0,0 +1,128 @@
...
+};
+const getOrderDateAndTime = () => {
+  const date = new Date();
+
+  const month = date.getMonth() + 1;
+  const day = date.getDate();
+  const year = date.getFullYear();
+
+  return {
+    orderdate: year * 10000 + month * 100 + day,
```
0Tendo, line 125: Just curious, why we have to multiply by 10000 and 100 here?

0Tendo (commented):
We discussed this at last weeks stand up, theres an issue with the seed data thats causing the seeded events to show a false number for 'Available Tickets' until after you edit the ticket. Not sure this would effect production, or necessarily in the scope of this ticket, but noting it as odd. 

before edit:
image

after edit:
image

server/src/controllers/orderController.service.ts
benfaustin95: The date is actually a foreign key to the Date table, so we have parse and then reformat the date to match the format of the foreign key.

benfaustin95:
> We discussed this at last weeks stand up, theres an issue with the seed data thats causing the seeded events to show a false number for 'Available Tickets' until after you edit the ticket. Not sure this would effect production, or necessarily in the scope of this ticket, but noting it as odd.
> 
> before edit: image
> 
> after edit: image

I added a quick update so now it should be on the first sale of the ticket that this fixes itself, but the problem itself should only ever exist with seed data as available seats is initialized to total seats when an event instance is created.

server/prisma/migrations/20231012180514_updated_ticketing_tables/migration.sql
```diff
@@ -0,0 +1,38 @@
...
+ALTER TABLE "singletickets" DROP COLUMN "eventticketid_fk";
+
+-- AddForeignKey
+ALTER TABLE "eventtickets" ADD CONSTRAINT "eventtickets_singleticket_fk_fkey" FOREIGN KEY ("singleticket_fk") REFERENCES "singletickets"("singleticketid") ON DELETE SET NULL ON UPDATE NO ACTION;
+
+-- AddForeignKey
+ALTER TABLE "orderitems" ADD CONSTRAINT "orderitems_orderid_fkey" FOREIGN KEY ("orderid_fk") REFERENCES "orders"("orderid") ON DELETE CASCADE ON UPDATE NO ACTION;
+
+-- AddForeignKey
+ALTER TABLE "singletickets" ADD CONSTRAINT "singletickets_orderitemid_fk_fkey" FOREIGN KEY ("orderitemid_fk") REFERENCES "orderitems"("orderitemid") ON DELETE CASCADE ON UPDATE NO ACTION;
```
HaliTran, line 38: My only concern is if any of this migration will have unintended effects on other parts of the app that we don't see. I know you pointed out the seed data being invalid which is okay b/c it can be fixed. Overall, it seems okay, just curious how you feel about it?

benfaustin95: The primary change I made was switching the ownership of the relationship between singltickets and eventickets, as far as I can tell no other code utilized the table so I think that should be fine. I also removed the purchased column from the eventickets table, for that I went through the repository and updated the locations where the field was used. Overall I think the changes should be good.

HaliTran: Okay that sounds good! Just wanted to make sure b/c there are a lot of relationships in the DB!

HaliTran (approved):
Seems like there is a merge conflict but once you resolve that, it looks good to me!

0Tendo (approved):

PR #395, opened 2023-10-12
TIC-77: Update swagger page and fix api for seasons

HaliTran:
### Description
- Updated the OpenAPI specification for the seasons swagger page API endpoint to include the new imageurl data column that was added. Specifically for the endpoints getSeason, getAllSeason, and updateSeason.
- Fixed the deleteSeason endpoint because previously, we could not delete a season since there are events that reference it. I ran a database migration to make the seasonID foreign key of the Events table optional so that we can now delete seasons even when there are events that reference them. This change will make the seasonID column for events NULL if a season that it references is deleted.

### Risks
Very minimal, although DB migrations always run the risk of having unseen consequences. However, considering  season was never utilized to begin, the migration shouldn't have any negative effects.

### Validation
Tested every API endpoint for seasons to ensure it does what it needs to do. Double checked the results in both pgAdmin and the swagger page.

### Issue
https://pdx-hkaus.atlassian.net/jira/software/projects/TIC/boards/1?selectedIssue=TIC-103

### Operating System
M2 Macbook

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -581,9 +581,6 @@ const createServer = async () => {
 
   let server;
 
-  console.log('process.env');
-  console.log(process.env);
-
   if (process.env.ENV === 'local') {
     const privateKey = fs.readFileSync('localhost-key.pem', 'utf8');
     const certificate = fs.readFileSync('localhost.pem', 'utf8');
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -207,6 +207,7 @@ const openApiSpec = swaggerJsdoc({
             name: {type: 'string'},
             startdate: {type: 'integer'},
             enddate: {type: 'integer'},
+            imageurl: {type: 'string'},
           },
         },
         SeasonTicket: {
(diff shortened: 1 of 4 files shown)


server/src/server.ts
```diff
@@ -510,20 +512,14 @@ const openApiSpec = swaggerJsdoc({
 const createServer = async () => {
```
HaliTran, line 512: Everything below this line is prettier formatting

server/src/server.ts
```diff
@@ -510,20 +512,14 @@ const openApiSpec = swaggerJsdoc({
...
   let envPath;
   if (process.env.ENV === 'local') {
-      envPath = path.join(__dirname, '../../.env');
-      // console.log("local")
-      // console.log(process.env)
+    envPath = path.join(__dirname, '../../.env');
   } else if (process.env.ENV === 'dev') {
-      envPath = path.join(__dirname, '../.env');
-      // console.log("dev")
-      // console.log(process.env)
```
HaliTran, line 519: Deleted unused code that was commented out.

server/src/server.ts
```diff
@@ -585,20 +581,19 @@ const createServer = async () => {
 
   let server;
 
-  console.log("process.env")
-  console.log(process.env)
+  console.log('process.env');
```
henrykaus, line 584: Could you remove these console.logs while you're at it

HaliTran: Done!

benfaustin95 (approved):

PR #400, opened 2023-10-17
TIC-77: Create select season and edit season page

HaliTran:
### Description
Added page to create and edit seasons. NOTE: There is currently no events being displayed when you click on a season as it is a work in progress. Please let me know if you have any questions, I tried my best to make the PR as small as possible.

### Risks
Minimal to no risk, this is a new feature so it doesn't conflict with the currently running application

### Validation
Manual UI test as well as checking pgAdmin and the swagger page to ensure database is creating and/or updating the seasons correctly.

### Issue
https://pdx-hkaus.atlassian.net/jira/software/projects/TIC/boards/1?selectedIssue=TIC-77

### Operating System
M2 MacBook


client/src/components/Ticketing/ticketingmanager/Season/components/SeasonEvents.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const SeasonEvents = () => {
+  return <h1>Season Events</h1>;
+};
+
+export default SeasonEvents;
```
HaliTran, line 7: Work in progress...

client/src/components/Ticketing/ticketingmanager/Season/components/utils/ViewSeasonInfo.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <button
+          className='bg-blue-500 hover:bg-blue-700 disabled:bg-gray-500 text-white font-bold py-2 px-10 rounded-xl'
+          onClick={() => setIsFormEditing(true)}
+        >
+          Edit
+        </button>
+      </section>
+
+      <section className='grid grid-cols-12'>
+        <article className='col-span-12 mb-5 text-center tab:text-start sm:col-span-6'>
```
henrykaus, line 48: Same here for `tab:col-span-6`

client/src/components/Ticketing/ticketingmanager/Season/components/utils/apiRequest.ts
```diff
@@ -0,0 +1,102 @@
...
+        method: 'PUT',
+        headers: {
+          'Content-Type': 'application/json',
+          'Authorization': `Bearer ${token}`,
+        },
+        body: JSON.stringify(reqBody),
+      },
+    );
+
+    if (!updateSeasonRes.ok) {
```
henrykaus, line 93: Is it possible for it not to throw an error, and also not be `ok`?

henrykaus (changes requested):
Left some comments, mostly minor changes. This is looking really good and intuitive. I did run into a couple issues though in testing. I found that when in the edit screen, if I clicked "Use Default Image", and then also cancelled my changes, the default image would still persist, and not the old image that was overwritten. I also see that there are some minor issues on mobile. It doesn't need to be perfect, but probably shouldn't have the buttons cascading wierdly.

image

client/src/components/Ticketing/ticketingmanager/Season/components/SeasonInfo.tsx
```diff
@@ -0,0 +1,188 @@
...
+              seasonId === 0
+                ? navigate('/ticketing/seasons/')
+                : setIsFormEditing(false);
+            }}
+          >
+            Cancel
+          </button>
+        </article>
+      </section>
+      <section className='grid grid-cols-12'>
```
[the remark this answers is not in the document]
HaliTran: Yeah that make sense, will do!

client/src/components/Ticketing/ticketingmanager/Season/components/utils/apiRequest.ts
HaliTran: The error is being thrown only if the HTTP status code isn't a success 200 level code, so if I'm understanding your question correctly, no its not possibly because if it is not ok, that means a 400 or 500 level status code was return which means an error has occurred.

henrykaus: Ok, well if that’s the case, I don’t think that if clause will ever be reached because it would have already been cause by the `except`

HaliTran: Sorry, I didn't explain that very well. If the status code response from the API is 400 or 500 which indicates an error (either b/c of the client or server), Javascript won't throw an error because technically since the API responded with a status code, it is considered a success. Therefore, if we didn't have this if statement to check if the status code was 200 and we got a 400 or 500 level status code, the catch clause won't ever be executed.

henrykaus: Oh ok, I see. Makes sense

henrykaus (approved):

PR #418, opened 2023-10-27
[TIC-111] Refund Page Rework (Front-end)

jonathantnguyen:
### Description
Admins are able to refund full orders.

Refund Process:
1. Input customer email address associated with the purchased order
2. A list of orders associated with the email will aggregate
3. Admin can select and refund the order and will be asked to confirm
4. Shows success/failure

image

image
image

### Risks
Not mobile friendly, but is desktop/tablet friendly

### Validation
Manual Testing

### Issue
[*Link to corresponding issue*](https://pdx-hkaus.atlassian.net/browse/TIC-111)

### Operating System
Win10/Chrome


client/src/components/Ticketing/ticketingmanager/Refund/RefundOrderTicket.tsx
```diff
@@ -1,42 +1,174 @@
...
+      price: 19.99,
+    },
+    {
+      name: 'John Doe',
+      date: '2023-10-25',
+      showings: 'Gone With The Wind',
+      orderID: '12345',
+      price: 14.99,
+    },
+  ];
```
HaliTran, line 68: Is this dummy data that will be removed later?

client/src/components/Ticketing/ticketingmanager/Refund/RefundOrderTicket.tsx
```diff
@@ -1,42 +1,174 @@
...
+                    ) : (
+                      <div>{instance.showings}</div>
+                    )}
+                  </div>
+                  <div className='row-start-1 justify-self-start pl-2 py-2 col-span-1'>
+                    {instance.price}
+                  </div>
+                  <div className='row-start-1 justify-self-start py-2 col-span-1'>
+                    <button
+                      onClick={async () => await onSubmit(instance.orderID)}
```
HaliTran, line 158: Async and await can be used here but it's not required because you declared the `onSubmit` function as async on line 17. So if you change this to just a regular arrow function like `() => onSubmit(instance.orderID)`, it'll work as intended as well. The `onSubmit` declared on line 17 will return a promise so it makes sense that would would do await onSubmit(...) here, but since you wrapped that in an async function, it returns another promise which isn't being used. So you can kinda skip the middle step and and remove the async await on this line. I hope that makes sense!

client/src/components/Ticketing/ticketingmanager/Refund/RefundOrderTicket.tsx
```diff
@@ -1,42 +1,174 @@
...
+                      type={'submit'}
+                      className='bg-red-700 hover:bg-red-800 text-white py-2 px-4 rounded'
+                    >
+                      Refund
```
HaliTran, line 162: When refunding the John Doe ticket, is it suppose to remove it from the table? When I click Refund, it says the refund was unsuccessful. However, I saw in the picture you provided in the description that it was removed for you? Just wanted some clarifications on that.

HaliTran (commented):
Left a couple of comments, but overall this looks great!

client/src/components/Ticketing/ticketingmanager/Refund/RefundOrderTicket.tsx
jonathantnguyen: Correct, dummy data will be removed since this is just front end development

client/src/components/Ticketing/ticketingmanager/Refund/RefundOrderTicket.tsx
jonathantnguyen: Currently not implemented, I'll see if this can be done front-end or requires an update/refresh to the db after refunding

benfaustin95: That is implemented on branch 95.

client/src/components/Ticketing/ticketingmanager/Refund/RefundOrderTicket.tsx
```diff
@@ -1,42 +1,174 @@
...
+              </div>
+            </label>
+          </div>
+          <div className='w-full min-w-min'>
+            <div className='grid grid-cols-5 gap-2 bg-gray-200 h-18 rounded-lg shadow-md px-2 mb-2 font-bold'>
+              <div className='row-start-1 justify-self-start py-2 col-span-1'>
+                  Name
+              </div>
+              <div className='row-start-1 justify-self-start py-2 col-span-1'>
+                <div className="border-l border-l-gray-500/50 pl-2">
```
henrykaus, line 117: I think you can also just combine these divs into one by having one element house all of those classes

client/src/components/Ticketing/ticketingmanager/Refund/RefundOrderTicket.tsx
```diff
@@ -1,42 +1,174 @@
...
+                </div>
+              </div>
+              <div className='row-start-1 justify-self-center py-2 col-span-1'></div>
             </div>
-            <button className='bg-red-700
-              hover:bg-red-500 text-white mt-5 py-2 px-4 rounded'>
-              Refund
-            </button>
+            {mappedInstances.length === 0 ? (
+              <div className="text-center text-gray-600">No current results</div>
```
henrykaus, line 134: This should be a `p` element since its sole purpose to display text

henrykaus (commented):
All some behind-the-scenes UI changes. The visuals look fantastic!

server/src/controllers/orderController.ts
```diff
@@ -45,6 +44,42 @@ orderController.post(
...
+          await ticketingWebhook(
+              prisma,
+              event.type,
+              object.payment_intent,
+              object.id,
+          );
+        // handle donation amount in metadata (will be a donation team task?)
+        } else if (metaData.sessionType === '__donation') {
+        // donation is handled
+        }
```
henrykaus, line 73: This should be brought up with the donation team

server/src/controllers/orderController.ts
```diff
@@ -187,6 +222,65 @@ orderController.get('/', async (req: Request, res: Response) => {
   }
 });
 
+
+orderController.get('/active', async (req: Request, res: Response) => {
```
henrykaus, line 226: We may not want to use `active` since we want to protect the meaning of active to be for events and seasons only. Might be better to use something else

server/src/controllers/orderController.ts
benfaustin95: I talked to them a couple weeks ago. Both teams have to use the same webhook api callback so the sessionType is being used to determine how to direct control flow between for donation or ticket transaction.

server/src/controllers/orderController.ts
benfaustin95: Sounds good, I ended up writing a different implementation of the route on branch 95 (refund backend). So we can just delete this route from the file if we want to.

henrykaus: Ok, lets delete this then

server/src/controllers/orderController.ts
henrykaus: Ok, thanks!

server/src/controllers/orderController.ts
benfaustin95: Fixed

henrykaus (approved):

PR #431, opened 2023-11-05
TIC-114 Front-End Active Events Component Rework

jonathantnguyen:
### Description
Switching to drop down selection to Group Toggle MUI design component that is it's own file for re-usability.

Before:
image

After:
image
image

### Risks
None currently, just front-end work.

### Validation
Manual Testing

### Issue
[*Link to corresponding issue*](https://pdx-hkaus.atlassian.net/browse/TIC-114)

### Operating System
Win10/Chrome


client/src/components/Ticketing/GroupeToggle.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
```
henrykaus, line 1: Usually we just want to import as little as possible, so we probably want to do the specific things we're importing, even though I know thats what the mui docs look like

client/src/components/Ticketing/GroupeToggle.tsx
```diff
@@ -0,0 +1,45 @@
...
+    setFilter(newFilterSetting);
+  };
+
+  return (
+    <ToggleButtonGroup
+      color='standard'
+      value={filterAs}
+      exclusive
+      onChange={handleFilterChange}
+      aria-label='filterSetting'
```
henrykaus, line 21: `aria-label`s are essentially supposed to be read outloud by a screen reader so it should be something understandable yet simple. I'd probably recommend something like 'activeness toggle filter'

client/src/components/Ticketing/GroupeToggle.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+import ToggleButton from '@mui/material/ToggleButton';
+import ToggleButtonGroup from '@mui/material/ToggleButtonGroup';
+
+const GroupToggle = () => {
```
henrykaus, line 5: Since this is specifically for toggling between activeness for ticketing things, I'd be a little more specific with the component name rather than a generic GroupToggle, and then make the filename match

client/src/components/Ticketing/GroupeToggle.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const [filterAs, setFilter] = React.useState('Active');
```
henrykaus, line 6: For consistency sake, the second item in the use state should be named the same except with a prepended `set`. So this can either be `[filterAs, setFilterAs]` or `[filter, setFilter]`

henrykaus: I also imagine that this should exactly match the value fields below, so `'active'`. This way `active` will be selected by default when the page loads

client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
```diff
@@ -2,6 +2,7 @@ import React, {useEffect, useState} from 'react';
 import {useNavigate} from 'react-router-dom';
 import {titleCase} from '../../../../utils/arrays';
 import {getSeasonImage, formatSeasonDate, SeasonImage} from './seasonUtils';
+import GroupToggle from '../../GroupeToggle';
```
0Tendo, line 5: Just curious, any reason for some being Group and some being Groupe?

client/src/components/Ticketing/GroupeToggle.tsx
henrykaus: This field would also be a good prop to add to this component. Something like `defaultValue` that could only be `active`, `inactive` or `all` and then gets thrown into the `useState` as the baseline

0Tendo (approved):
I like this change. Good idea making it a separate component. We should incorporate this into the doorlist once we get the backend working.

henrykaus (commented):
Just a couple comments, it overall looks good and functions great. My only stylistic suggestion would be to make the text a little bolder, it looks a little thin at the moment and makes the gray text hard to read. I think grey is good, but the text should be a little bolder if possible

client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
jonathantnguyen: Thanks for catching the typo. Fixed.

client/src/components/Ticketing/GroupeToggle.tsx
jonathantnguyen: On line 6, I use `const [filterAs, setFilter] = React.useState('Active');`. Is this still unneeded? I result in code being underlined in red.

jonathantnguyen: On line 6, I use ` const [filterAs, setFilter] = React.useState('Active');`. Is it still unneeded? Currently, removing the import results in code being underlined in red.

henrykaus: Well, you wouldn't want to remove the import but something more along the lines of `import {useState} from 'react';`

henrykaus: And that import statement would also need to include anything else that is imported from `'react'` like MouseEvent

client/src/components/Ticketing/GroupeToggle.tsx
jonathantnguyen: added/fixed

jonathantnguyen:
New changes include:

- FontWeight: Bold
- Added defaultValue to component & associated files which need to include "defaultValue"
- Renamed files, component name, and labels.

client/src/components/Ticketing/GroupeToggle.tsx
jonathantnguyen: fixed

client/src/components/Ticketing/GroupToggle.tsx
```diff
@@ -44,5 +45,9 @@ const ShowingActivenessToggle =(defaultValue) => {
   );
 };
 
+ShowingActivenessToggle.propTypes = {
+  defaultValue: PropTypes.string.isRequired,
+};
+
```
henrykaus, line 51: I love this library so I'm glad you're using it

henrykaus (approved):
Thanks for implementing all of the changes! This component looks good

PR #437, opened 2023-11-10
TIC-124: Migrate Season page to utilized v2 events API instead of v1

HaliTran:
### Description
Update Season page to utilized most-up-date events API that was changed recently. 

### Risks
Little to minimal risks, simply updated which API was being used and removed code that was no longer needed

### Validation
Test in the UI to ensure everything still functions as normal and double check database for changes.

### Issue
https://pdx-hkaus.atlassian.net/jira/software/projects/TIC/boards/1?selectedIssue=TIC-124

### Operating System
M2 Macbook

### DIFF
--- a/client/src/components/Ticketing/ticketingmanager/Season/components/SeasonEvents.tsx
+++ b/client/src/components/Ticketing/ticketingmanager/Season/components/SeasonEvents.tsx
@@ -41,7 +41,7 @@ const SeasonEvents = (props: SeasonEventsProps) => {
 
     eventToUpdate = {
       ...eventToUpdate,
-      seasonid_fk: undefined,
+      seasonid_fk: null,
     };
 
     const updateEventCall = await updateEventSeason(eventToUpdate, token);
--- a/client/src/components/Ticketing/ticketingmanager/Season/components/SeasonEvents.tsx
+++ b/client/src/components/Ticketing/ticketingmanager/Season/components/SeasonEvents.tsx
@@ -19,11 +19,12 @@ const SeasonEvents = (props: SeasonEventsProps) => {
   const handleGetAllEvents = async () => {
     const allEvents = await getAllEvents(token);
     if (allEvents) {
-      const {data} = allEvents;
-      const eventsInSeason = data.filter(
-        (event) => event.seasonid === seasonId,
+      const eventsInSeason = allEvents.filter(
+        (event) => event.seasonid_fk === seasonId,
       );
-      const eventsNotInSeason = data.filter((event) => event.seasonid === null);
+      const eventsNotInSeason = allEvents.filter(
+        (event) => event.seasonid_fk === null,
+      );
 
       setAllEventInfo(eventsInSeason);
       setEventsNotInSeason(eventsNotInSeason);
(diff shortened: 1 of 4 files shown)


server/src/controllers/eventController.ts
```diff
@@ -743,7 +743,7 @@ eventController.put('/', async (req: Request, res: Response) => {
         eventid: Number(req.body.eventid),
       },
       data: {
-        seasonid_fk: Number(req.body.seasonid_fk),
```
HaliTran, line 746: @benfaustin95 Converting to a Number caused the update events endpoint to throw an error when a NULL value was used for season id. Not sure if there are any major side effects with this change, will like a second opinion.

client/src/components/Ticketing/ticketingmanager/Season/components/SeasonEvents.tsx
```diff
@@ -19,11 +19,12 @@ const SeasonEvents = (props: SeasonEventsProps) => {
   const handleGetAllEvents = async () => {
     const allEvents = await getAllEvents(token);
     if (allEvents) {
-      const {data} = allEvents;
-      const eventsInSeason = data.filter(
-        (event) => event.seasonid === seasonId,
+      const eventsInSeason = allEvents.filter(
+        (event) => event.seasonid_fk === seasonId,
+      );
+      const eventsNotInSeason = allEvents.filter(
```
henrykaus, line 25: For clarity sake could we rename this to `eventsNotInASeason`. At the moment, it looks like it means all events that are not in this season

client/src/components/Ticketing/ticketingmanager/Season/components/SeasonInfo.tsx
```diff
@@ -228,21 +228,10 @@ const SeasonInfo = (props: SeasonProps) => {
...
-                id='activeSeason'
-                name='activeSeason'
-              />
-              <label className='text-base' htmlFor='activeSeason'>
-                Active
-              </label>
+
+              <FormControlLabel control={<Checkbox />} label='Active' />
             </div>
+            <div className='checkbox-2'></div>
```
henrykaus, line 234: Is this `div` necessary?

client/src/components/Ticketing/ticketingmanager/Season/components/SeasonInfo.tsx
```diff
@@ -213,13 +214,12 @@ const SeasonInfo = (props: SeasonProps) => {
             />
           </label>
           <div id='form-checkboxes' className='flex gap-7'>
-            <div className='checkbox-2'>
-              <input
-                type='checkbox'
-                id='defaultImage'
-                name='defaultImage'
-                className='mr-2'
+            <div className='checkboxes'>
```
henrykaus, line 217: I don't think this class does anything, I would instead maybe make this a flex object and do styling on this element rather using `sx` in the innner elements. On mobile, the checkboxes are a little weird

henrykaus (commented):
This wasn't you, but in the svg elements, could you make sure we're using camel case for all props (except aria or testing-related props), I noticed some of the `svg` props are not react friendly. Otherwise everything works great and this is significantly less code

client/src/components/Ticketing/ticketingmanager/Season/components/SeasonInfo.tsx
HaliTran: No it's not, removed it shall be!

client/src/components/Ticketing/ticketingmanager/Season/components/SeasonInfo.tsx
henrykaus: image

HaliTran: I'm actually going to revert this back to the old checkbox and make the change in the next PR. I was working on the seasons active button but had to update the api being used for it to work properly and forgot to stash my changes.

henrykaus: Sounds good 👍🏻

client/src/components/Ticketing/ticketingmanager/Season/components/SeasonEvents.tsx
HaliTran: That makes sense, will do!

HaliTran: Now that I took a look, it actually does refer to all events not in the current season, so the naming scheme makes sense. Since I store all the events in `eventsInSeason` variable, everything else is not in the season

henrykaus: But isn't it checking for all events that have a `null` foreign key. That means that events that are in _different_ seasons are also not included in this var

HaliTran: Hmm yeah you're right in that sense. I guess I just grouped all the events that were not in the current season (whether it was null or a part of another season) together. I'll changed that to be more clear!

henrykaus: I guess if it makes more sense you could do something like: `eventsNotInAnySeason`

benfaustin95 (approved):

PR #438, opened 2023-11-11
DEP-81: use a single pipeline to run E2E tests and deploy to Cloud Run

jpb7:
### Description
There are three basic problems with our current E2E testing on deployed versions of WonderTix:
1. We have to rely on a naive `sleep` command and hope the deployed client spins up in time.
2. We're unable to automate any kind of corrective action to be taken if the deployed version fails the E2E tests. Whatever we send up to GCP goes live, whether it works or not.
3. We're using three different triggers to test, build, and deploy: one of them from GitHub Actions and two from Cloud Build.

This branch consolidates our client, server, and E2E workflows into a single pipeline that interfaces tightly with Cloud Run. It allows us to use a single Cloud Build trigger to run tests and it will abort the deployment if tests fail. It also uses a Playwright script to programmatically request a page load from the deployed client, pausing the workflow until the client has verifiably started accepting requests.

image

Most importantly, we're able to run E2E tests in a live environment without interrupting our current deployments, which continue to serve traffic undisturbed. The revisions being tested on Cloud Run receive no traffic and are effectively invisible to users.

image

Here are the basic steps:
- Build, push, and deploy test revisions of both client and server, ensuring they receive no traffic.
- Run E2E tests on those revisions in the live environment.
- If all tests pass, deploy the new client and server and send them traffic; otherwise, abort the workflow.

There are two additional changes worth noting:
1. I've removed the unused `terraform` directory from the repo.
2. I've refactored our `cloudbuild.yaml` files to use Bash scripts for longer or more complex steps.

These Bash scripts are located in the new `scripts` directory in the top level of the repo. They use a naming convention that is a bit terse: for example, instead of using script names like `deploy-client.sh` or `deploy-server.sh`, I've created a `deploy` sub-directory and named the scripts simply `client.sh` or `server.sh`. The meaning is clear in context: our workflow steps will specify a script using `scripts/deploy/client.sh` or `/scripts/deploy/server.sh`.

image

Our `client/cloudbuild.yaml` and `server/cloudbuild.yaml` are now using these Bash scripts for certain steps. Each of those workflows can still run via our manual triggers in Cloud Build.

Finally, we will need to make changes both in the GitHub repo and on GCP after merging this branch. In GitHub, we should remove our current deployment E2E testing as a required check; on GCP, we'll need to disable the existing triggers and adjust the trigger corresponding to this unified workflow so that it runs automatically on pushes to the `main` branch.

image

### Risks
- **Important note**: since we're still ironing out one or two of our E2E tests, this current pipeline will run in its entirety even if the tests fail. In other words, it will report success as long as the deployment itself is successful. Once we have fixed the relevant E2E tests, we can un-comment a single line `exit 1` in `scripts/test/client.sh` that will cause the workflow to accurately report success or failure, and to gatekeep deployment accordingly.
```
# Run Playwright tests
if ! npm run test:playwright; then
  echo "Tests failed."
  # Uncomment the line below to fail the script when tests fail
  # exit 1
  echo "Ignoring failed tests for now."
else
  echo "All tests passed."
fi
```
- It's more of a weakness than a risk, but this pipeline is kind of a beast: it currently takes around 20 minutes to run from start to finish. I have a couple of follow-up tickets to this one that have a good chance of cutting that time down to around 10 minutes.

image

- I don't think removing the `terraform` directory is actually a risk, but because it kind of _feels_ like one, I'll just add here that I backed it up on my local machine out of an abundance of caution.
- It isn't actually any different from the current state of `main`, but sometimes if you navigate to the client at the wrong moment and get really lucky, you'll see the main page without its proper background image. This only occurs for a period of 30 seconds when the new client begins to receive traffic, and even then, it only happens if you're part of the 5% of traffic that goes to that revision. I'm hoping that my next couple of PRs, along with smoothing down some rough edges in Auth0, will solve this problem along with a few others.

image

### Validation
I've tested this _a lot_. Doesn't mean it's perfect or that I didn't miss anything, but you can look in our build history on GCP and check out all the instances of `DEP-81` for more details. Here's a link to the most recent build log:
- https://console.cloud.google.com/cloud-build/builds;region=us-west2/2df15a4e-f665-432d-9bfb-73f02fc15768?project=wondertix-app

Also, I'm attaching a PDF containing a step-by-step demonstration of how this pipeline works in Cloud Build and Cloud Run. Seeing the screenshots can make it a lot clearer than trying to read and parse the YAML (or this PR description, perhaps).

- [DEP-81-demo.pdf](https://github.com/WonderTix/WonderTix/files/13324574/DEP-81-demo.pdf)

### Issue
- Jira ticket: [DEP-81: Run deployment E2E tests using Cloud Build](https://wondertix.atlassian.net/browse/DEP-81)

### Operating System
- Ubuntu 22.04


client/playwright.config.ts
jpb7, line 1: These additions allow us to use simple text reports instead of the full, 100+ MB records currently being generated in GitHub Actions.

scripts/test/client.sh
jpb7, line 1: Line `40` here is the one we can un-comment to cause failing tests to abort the workflow.

server/cloudbuild.yaml
jpb7, line 1: The `send-traffic` step here ensures that new server revisions will automatically receive traffic.

Sometimes when I'm actively tinkering with our workflows, the most recent revision will be dead-in-the-water and not receiving traffic; if someone merges a PR at that moment, the new server will inherit the traffic settings of that dead revision and it won't receive any traffic.

scripts/traffic/client.sh
jpb7, line 1: This script tries valiantly to avoid cold starts (and it mostly succeeds), but the `tl;dr` is this: optimizing our client image is the best way out of this mess.

cloudbuild.yaml
jpb7, line 1: I forgot to mention it in the PR description, but the `id` and `waitFor` fields here allow us not only to specify execution order, but also to run steps concurrently. For example, lines `6` and `53` indicate that the server and client will both run as soon as the workflow starts. In other words, the server and client will build, push, and deploy concurrently.

jtng3 (approved):

jtng3 (approved):

PR #439, opened 2023-11-11
[TIC-116] Fix Pay What You Can

henrykaus:
### Description
When adding Pay What You Can tickets to an event showing, the following scenarios have been fixed:
1) Adding any amount of tickets to the cart with the set price will keep the set price in the cart
2) Increasing or decreasing the amount of tickets from the cart does not affect the price
3) When adding more tickets to the cart from the ticket picker, the Pay What You Can price will change to the new set price
4) Buying Pay What You Can tickets reflects within Stripe by using the `payWhatPrice` value
5) Can no longer attempt to buy negative price for ticket type
6) Pay What You Can price is now rounded to 2 decimal places en route to cart

**Other changes:** 
1) Updated phone number validation to make the `+` optional if providing a region code
2) Fixed bug where if you were to add 2 or more ticket types for the same showing to the cart, and wanted to remove all of one, it would remove all for that showing

### Risks
I tested it as much as a can but there may be a scenario that I missed.

### Validation
Added Pay What You Can ticket type to an event showing. Increased and decreased the amount of tickets to purchase in various ways. Went through purchase process and the correct price was reflected in Stripe. Checked the DB to ensure that the order total and individual ticket prices were correct.

### Issue
[TIC-116](https://pdx-hkaus.atlassian.net/browse/TIC-116)

### Operating System
Windows 11 / Chrome

[TIC-116]: https://pdx-hkaus.atlassian.net/browse/TIC-116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henrykaus (commented):
**_Validation:_**
image
image
image
image
image
image

jonathantnguyen (commented):
While I have looked at it, and tried to break it, everything works, but probably best if another team member looks over this as well.

benfaustin95 (approved):

PR #440, opened 2023-11-11
[TIC-120] Group Toggle for Seasons Page with new API Call

jonathantnguyen:
### Description
Created a new API call off of Prisma, `/season/list`, which list the seasons and events within.

Group Toggle is similar to Events Group Toggle where the options are Active/Inactive/All

Criteria:
Active: All events within the season must be active.
Inactive: At least 1 event is inactive in a season.
All: Combination display

Display is by most recent creation.

image

### Risks
None known, API call is strict.

Documentation Block & Code in the `seasonController.ts` must be thoroughly reviewed due to inexperience with API.

### Validation
Manual Testing:
image

Current seeded data show only known active season is "Easter Egg Hunt", where all events are active.

* There are 7 Seasons. For each of all inactive, all events within are seeded as inactive.

### Issue
[*Link to corresponding issue*](https://pdx-hkaus.atlassian.net/browse/TIC-120)

### Operating System
Win10/Chrome

### DIFF
--- a/client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
+++ b/client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
@@ -28,32 +28,57 @@ const SeasonInstancesPage = (props: SeasonInstancesProp) => {
   const {token} = props;
   const [filterSetting, setFilterSetting] = useState('active');
   const [seasonData, setSeasonData] = useState([]);
-
+  const [allSeasonData, setAllSeasonData] = useState([]);
+  const [inactiveData, setInactiveData] = useState([]);
+  const [activeData, setActiveData] = useState([]);
 
   useEffect(() => {
-    let apiURL;
-
-    if (filterSetting === 'active') {
-      apiURL = process.env.REACT_APP_API_2_URL + '/season/active';
-    } else if (filterSetting === 'inactive') {
-      apiURL = process.env.REACT_APP_API_2_URL + '/season/inactive';
-    } else {
-      apiURL = process.env.REACT_APP_API_2_URL + '/season';
-    }
-
-    fetch(apiURL, {
+    fetch(process.env.REACT_APP_API_2_URL + '/season/list', {
       headers: {
         'Authorization': `Bearer ${token}`,
       },
     })
     .then((response) => response.json())
     .then((data) => {
-      console.log('Received Season Data:', data);
-      setSeasonData([...data.toSorted()].reverse());
+      const inactiveSeasons = [];
+      const activeSeasons = [];
+      const allSeasons = [];
+
+      data.forEach((season) => {
+        const allEventsActive = season.events.every((event) => (event.active));
+
+        if (allEventsActive) {
+          activeSeasons.push(season);
+          allSeasons.push(season);
+        } else {
+          inactiveSeasons.push(season);
+          allSeasons.push(season);
+        }
+      });
+      // Set active/inactive/all data arrays
+      setInactiveData(inactiveSeasons);
+      setActiveData(activeSeasons);
+      setAllSeasonData(allSeasons);
+      // Set Default "Active"
+      setSeasonData(activeSeasons);
     })
     .catch((error) => {
       console.error('Error Fetching Season Data:', error);
     });
+  }, []);
+
+  // Group Toggle Display Changes
+  useEffect(() => {
+    if (filterSetting === 'active') {
+      // Active Data Only
+      setSeasonData([...activeData].sort((a, b) => b.seasonId - a.seasonId));
+    } else if (filterSetting === 'inactive') {
+      // Inactive Data Only
+      setSeasonData([...inactiveData].sort((a, b) => b.seasonId - a.seasonId));
+    } else {
+      // All Seasons
+      setSeasonData([...allSeasonData].sort((a, b) => b.seasonId - a.seasonId));
+    }
   }, [filterSetting]);
 
   return (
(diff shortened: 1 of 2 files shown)


client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
```diff
@@ -26,45 +26,60 @@ interface SeasonInstancesProp {
...
+  useEffect(() => {
+    if (filterSetting === 'active') {
+      // Active Data Only
+      setSeasonData([...activeData].sort((a, b) => b.seasonId - a.seasonId));
+    } else if (filterSetting === 'inactive') {
+      // Inactive Data Only
+      setSeasonData([...inactiveData].sort((a, b) => b.seasonId - a.seasonId));
+    } else {
+      // All Seasons
+      setSeasonData([...allSeasonData].sort((a, b) => b.seasonId - a.seasonId));
```
henrykaus, line 80: To reduce some of the operations (the spreading and sorting each time the toggle is pressed), you could move the sorts to lines 59-61.

server/src/controllers/seasonController.ts
```diff
@@ -145,6 +145,96 @@ seasonController.get('/', async (req: Request, res: Response) => {
...
 
+/**
+ * Retrieves a list of seasons with their associated events.
+ *
+ * @swagger
+ * /2/season/list:
+ *   get:
+ *     summary: Get all seasons with events
+ *     tags:
+ *     - Season
```
henrykaus, line 156: The tag should be `New season` so that it is in the same place as the rest of the season APIs
image

server/src/controllers/seasonController.ts
```diff
@@ -145,6 +145,96 @@ seasonController.get('/', async (req: Request, res: Response) => {
...
+        seasonName: season.name,
+
+        events: season.events.map((event) => {
+          return {
+            eventId: event.eventid,
+            eventName: event.eventname,
+          };
+        }),
+      };
+    });
```
henrykaus, line 223: I personally think it would probably be better if we were returning everything from the above query because these APIs should be built with the possibility of them being used in the future, and future uses may want to use more of the data beside just these 4 values.

client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
```diff
@@ -26,45 +26,60 @@ interface SeasonInstancesProp {
...
-    }
-  };
+      data.forEach((season) => {
+        const allEventsActive = season.events.every((event) => (event.active));
 
-  useEffect(() => {
-    getAllSeasons();
+        if (allEventsActive) {
+          activeSeasons.push(season);
+          allSeasons.push(season);
```
henrykaus, line 52: Up to you, but you could put the call to `allSeasons.push(season)` after the `if-else`

henrykaus (commented):
This might just be a me thing, but I can't actually get this to work on my end, is this working on your machine?

server/src/controllers/seasonController.ts
jonathantnguyen: fixed

client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
jonathantnguyen: refactored

server/src/controllers/seasonController.ts
jonathantnguyen: fixed

client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
jonathantnguyen: refactored

jonathantnguyen:
Hold off on the review, something broke.

server/src/controllers/seasonController.ts
```diff
@@ -145,6 +145,78 @@ seasonController.get('/', async (req: Request, res: Response) => {
...
+ *                     description: An array of events associated with the season.
+ *                     items:
+ *                       type: object
+ *                       properties:
+ *                         eventId:
+ *                           type: number
+ *                           description: The unique identifier of the event.
+ *                         eventName:
+ *                           type: string
+ *                           description: The name of the event.
```
henrykaus, line 184: You'll probably have to update this now that you've changed what the API returns, but this block is very handy to have when trying to look for APIs to use

client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
```diff
@@ -26,45 +26,64 @@ interface SeasonInstancesProp {
...
 
-  useEffect(() => {
-    getAllSeasons();
+        // Common operation
+        allSeasons.push(season);
+      });
+      // Sorting
+      inactiveSeasons.sort((a, b) => b.seasonId - a.seasonId);
+      activeSeasons.sort((a, b) => b.seasonId - a.seasonId);
+      allSeasons.sort((a, b) => b.seasonId - a.seasonId);
```
henrykaus, line 62: It looks like they might be sorted the reverse way, this might be because `seasonId` might be mismatched to what the API returns? Perhaps it should be `seasonid`.

henrykaus (changes requested):

client/src/components/Ticketing/ticketingmanager/Season/SeasonInstancesPage.tsx
jonathantnguyen: thanks, fixed

server/src/controllers/seasonController.ts
jonathantnguyen: updated

jonathantnguyen:
image

henrykaus (approved):
Looks and works great, thanks for making all the changes!

PR #454, opened 2023-11-14
[TIC-113] Ticket Type Page Updates

0Tendo:
### Description
1. Added column to display the concession price for tickets.
2. Removed the ability to delete ticket type 1 - All tickets pull from the ticket type 1 pool (General Admission - Adult by defualt). If this ticket type is deleted it will break a lot of other components.
3. Replaced the pop up message with the pre-defined PopUp component

image

### Risks
Other pages require dynamic ticket type implementation.

### Validation
Manual testing

### Issue
N/A

### Operating System
Windows 11


client/src/components/Ticketing/ticketingmanager/TicketTypes/tickettypes.tsx
```diff
@@ -36,39 +33,52 @@ const TicketTypes = () => {
...
-            () => {
-              setConfirmDeletePrompt(true);
-              setCellData(cell);
-            }
-          } />
-        );
+        // Check if the id of the row is 1
+        if (cell.row.id === 1) {
+          // Return a "Default" tag for id 1
+          return <Chip label="Default" color="primary" />;
```
henrykaus, line 51: At the moment, it kind of looks like the default `Chip` is a clickable button, so it might be better to make it have a `default` color and then possibly a `size='small'` just so it has a visual cue that it there isn't any functionality for that.

henrykaus (commented):
Just a visual tweak to the `Chip`. Everything else looks good, thanks for removing all of the logs and simplifying how deletion is handled :)

client/src/components/Ticketing/ticketingmanager/TicketTypes/tickettypes.tsx
0Tendo: good idea. made the change.

henrykaus (changes requested):
The visual changes look great, but it doesn't look like changes to concessions actually persist if you were to just change the concessions price, I imagine it would just be another condition in the `handleEditType` method

henrykaus (approved):
Thank you!

PR #464, opened 2023-11-17
TIC-122

alexfevery01:
### Description
New Event page Design

### Risks
Event page may have broken functionality

### Validation
testing

### Issue
[*Link to corresponding issue*](https://pdx-hkaus.atlassian.net/browse/TIC-122)

### Operating System
Windows


henrykaus (commented):
This is looking really good. There were a couple issues I was noticing when I was testing this branch on my own:

(1) On a mobile size, the buttons overflow their container, I imagine you could have them wrap onto the next line at smaller sizes

image

(2) When you set the image to be the default one and then save. When trying to change the image again, it doesn't let you. I imagine this has to do with the old way of restricting user input when the default image is set.

image

(3) When selecting a season, if the seasons could be sorted by `seasonid` descending, that would be more appropriate. This would make it so that the newest created seasons show up at the top, and older ones at the bottom.

henrykaus (approved):
Thanks for making all the changes and fixes!! I was able to modify the tests to work with your changes

HaliTran (approved):

PR #471, opened 2023-11-20
Dep 73 test purchase and cart functions

snorq2:
### Description
Updated:
Main Page object model to include Stripe and Cart processing elements.
Door List object model to search and check for patrons.

Added:
Contacts object model to search for and validate patrons.
Interfaces for Customers and Credit Card info to ConstsPackage

Removed 'add tickets' as a test, replaced by the below tests:

'check cart after ticket add' - Adds tickets from the main page and ensures they're in the card with the proper information.
'check contact is added after order' - Orders a ticket and ensures the patron is on the Contacts page with the correct information.
'check order on door list' - Orders a ticket and ensures the name and quantity appears on the Door List
'check ticket inc/dec in cart' - Adds tickets from the main page to the card and checks quantity and ticket total while increasing/decreasing the ticket count.  DOES NOT verify the overall subtotal - expect this to be a different test with multiple tickets.
'check order accommodations' - Similar to 'check contact is added after order' except uses a customer with an accommodation.

### Risks
The tests that include Stripe in their flow tend to be flaky.  The Stripe interface is slow - two common issues are a pop-up generating semi-randomly, and the transaction processing taking an extended period of time.  Waits have been added to handle such situations, but this only reduced the occurrence of test failure.  Repeating tests typically allowed them to pass.

### Validation
Ran individual and batch tests against all three supported browsers.

### Issue
https://wondertix.atlassian.net/browse/DEP-73?atlOrigin=eyJpIjoiMzJiMTEzY2NjMzJlNDk5NmIwN2JmZDVhNmY5MTNmOTgiLCJwIjoiaiJ9

### Operating System
Windows 11


jtng3 (approved):

client/e2e/testData/ConstsPackage.ts
```diff
@@ -8,66 +19,161 @@ export interface EventsInfo {
...
 export const EventsInfo4: EventsInfo = {
-  eventName: "The Crucible",
-  eventDescription: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
-  eventURL: "https://upload.wikimedia.org/wikipedia/en/7/75/Cruciblecover.jpg",
-  eventFullName: "The Crucible Playbill The Crucible Description Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
+  eventName: 'The Crucible',
+  eventDescription: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+  eventURL: 'https://upload.wikimedia.org/wikipedia/en/7/75/Cruciblecover.jpg',
+  eventFullName: 'The Crucible Playbill The Crucible Description Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+};
```
henrykaus, line 47: It might be a good idea to have a different event from the seed data so we can distinguish them in case one ends up being left around in dev

jtng3: good idea!

jtng3: sawyers next ticket will be to add uuid's to all test data for uniqueness. but henry's suggestion will indeed help troubleshoot any left over test data if they occur.

snorq2: This is actually my next project.  I plan to convert the event interface into a class and give it the ability to append a UID to the event name.

PR #473, opened 2023-11-20
Remove module import that is not being used.

0Tendo:
### Description
removed module import that was not being used and was causing E2E tests to fail.

### Risks
N/A

### Validation
E2E tests pass

### Issue
N/A

### Operating System
Windows


henrykaus (approved):
Merge once tests pass

PR #476, opened 2023-11-20
PPH database migration, updated with current branch

dstad418:
**Description:**

Added DB models for the PPH Salesforce integration, seed scripts for those tables and updated seeding routines. Re-added the original Wondertix models to ensure API stability.

Per Johnny:
"invite me, Eben, Ben Williams, Justin, Henry, Jaeger and Adam."

**Risks:**

Some of the seeding routines will compare the data given to it, and if no relation can be made, it will skip over that instance of data. Johnny approved this step, as one-sided relations broke the seeding process- changing the fields to optional or not requiring a relation didn't fix this.
Because of this, dataset validation and integrity is key to this working appropriately. At present, a lot of the DEV seed code isn't linked appropriately to the other tables' relations, which leads to a lot of seeded data being skipped over. Production tables have a better success rate.

**Validation:**

I've had to beef up the seeding scripts to sanitize data as it came in. Salesforce had a lot of mixed data types in any given field (I suspect all of their data is sanitized to strings), or dangling nodes of data that were probably assigned to a catch-all account.

Over several days I added changes to the seeding scripts to make them more robust. I checked it with the deployment tables, the dev tables, and instances of custom data to ensure they worked.

**Operating System:**

Mac, M2-series CPU


dstad418:
Hey all, Johnny wanted me to re-push with a more updated version of Main. 
Made the changes, confirms it works, it's here for review!

server/package.json
```diff
@@ -21,7 +22,7 @@
     "npm": ">=8.0.0"
   },
   "dependencies": {
-    "@prisma/client": "^5.5.2",
+    "@prisma/client": "^5.5.0",
```
henrykaus, line 25: Is there any particular reason that we're downgrading the prisma client (this and the one on line 63)?

henrykaus:
> Hey all, Johnny wanted me to re-push with a more updated version of Main. Made the changes, confirms it works, it's here for review!

Thanks for getting the latest version of `main` in this PR!

server/package.json
dstad418: Hmmm, I don't know how that change happened- must have been an accident. The only thing I meant to change in that file was adding the deployment script expand on nodes.js's default memory limit of 1.5gb. 
Good catch!

dstad418: Fixed in most recent push!

xiivi (approved):

jtng3 (approved):

ben16williams:
> At present, a lot of the DEV seed code isn't linked appropriately to the other tables' relations, which leads to a lot of seeded data being skipped over. Production tables have a better success rate.

Does this mean that you've tried with real patron manager data? I'd be curious what the success rate was. To your point about data integrity/validation, it sounds like we might have to manually add ID values to foreign key vales in some instances so that some items don't get skipped?

server/prisma/src/seedPPH_Contacts.ts
```diff
@@ -0,0 +1,116 @@
...
+          where: {account_id: item.account_id},
+        });
+
+        if (!accountExists) {
+          missingAccountsCount++; // Increment missing accounts counter
+          canInsert = false;
+
+          if (process.env.ENV === 'local') {
+            console.log(`Skipping contact with ID ${item.contact_id} due to non-existing account with ID ${item.account_id}`);
+          }
```
ben16williams, line 38: Any reason this is only logged when seeding locally? If it's matter of privacy, I'd be less concerned about logging ID values as opposed to PII, especially since presumably these are all logged in GCP somewhere which has restricted access. This does seem like something that would be nice to always have logged (same for the rest of the seedPPH files that have this kind of logic)

ben16williams (approved):
One comment about logging, not a blocker, otherwise LGTM. That's a lot of lines! Happy to reapprove if you have to make changes to get GHA to pass

dstad418:
> > At present, a lot of the DEV seed code isn't linked appropriately to the other tables' relations, which leads to a lot of seeded data being skipped over. Production tables have a better success rate.
> 
> Does this mean that you've tried with real patron manager data? I'd be curious what the success rate was. To your point about data integrity/validation, it sounds like we might have to manually add ID values to foreign key vales in some instances so that some items don't get skipped?

Hey Ben, 
I did try with the actual dev data, which had a higher success rate. I'll look into getting you a %. 
I think there's some dataset relation errors in the dev data, but it's only a few tables that have this issue.

server/prisma/src/seedPPH_Contacts.ts
dstad418: This was a request from Jaeger's team, as before it was looping over every skipped instance and console logging for each instance (For example, if a instance of data couldn't make a relation to accounts, contacts and opportunities, there would be three lines output for one object). 
They wanted a summary in lieu of all of that output. 

_Otherwise_, the hold up here looks like the Events page on the client side of things. I'll look into contacting the team who is handling this.

henrykaus (approved):

jgreever (approved):
Initial testing and looking around, everything works as expected. I did have to do the following on my system: `docker system prune --all --force` just to clean the entire docker system, then after doing the standard `npm install...` and `mkcert localhost` stuff in the directories, I switched to the server directory after doing a `docker compose up -d`. In that directory, I waited till the system was up and then did `npx prisma generate` then `npx prisma migrate dev`. From there, I didn't have any complaints from my IDE about tables being missing and the `npx prisma studio` worked and I could debug and run queries without having to use swagger directly.

It looks good, and I merged it with a few other branches locally just to test for some possible issues, and again, outside of sql queries needing to be re-written by the respective teams, all looks great! Nice work!

PR #478, opened 2023-11-21
TIC-134-A: adding SeasonTicketTypePriceDefault table and APIS

benfaustin95:
### Description
Adding a table and API calls as needed to modify ticket type prices by season ID.

Only includes backend functionality for the new table and does not include integration with the current schema.

Functionality includes:
- PUT: Accepts an array of season ticket type prices and adds/updates/deletes accordingly.
- GET: Retrieves all season ticket type prices linked to a specific season ID.

### Risks
NA

### Validation
Manual testing on Swagger page.

### Issue
[Jira Ticket](https://pdx-hkaus.atlassian.net/jira/software/projects/TIC/boards/1?selectedIssue=TIC-135)

### Operating System
M1 Mac.

### DIFF
--- a/server/prisma/schema.prisma
+++ b/server/prisma/schema.prisma
@@ -194,14 +194,15 @@ model savedreports {
 
 /// a season is a collection of events over a period of time (e.g. a summer season)
 model seasons {
-  seasonid                     Int      @id @default(autoincrement())
-  name                         String   @db.VarChar(100)
-  startdate                    Int
-  enddate                      Int
-  imageurl                     String?  @db.VarChar(255)
-  events                       events[]
-  date_seasons_enddateTodate   date     @relation("seasons_enddateTodate", fields: [enddate], references: [dateid], onDelete: NoAction, onUpdate: NoAction)
-  date_seasons_startdateTodate date     @relation("seasons_startdateTodate", fields: [startdate], references: [dateid], onDelete: NoAction, onUpdate: NoAction)
+  seasonid                      Int                            @id @default(autoincrement())
+  name                          String                         @db.VarChar(100)
+  startdate                     Int
+  enddate                       Int
+  imageurl                      String?                        @db.VarChar(255)
+  events                        events[]
+  date_seasons_enddateTodate    date                           @relation("seasons_enddateTodate", fields: [enddate], references: [dateid], onDelete: NoAction, onUpdate: NoAction)
+  date_seasons_startdateTodate  date                           @relation("seasons_startdateTodate", fields: [startdate], references: [dateid], onDelete: NoAction, onUpdate: NoAction)
+  seasontickettypepricedefaults seasontickettypepricedefault[]
 }
 
 /// a season ticket is a ticket that is not part of a bulk package of tickets during a season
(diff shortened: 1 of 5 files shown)


server/prisma/schema.prisma
```diff
@@ -230,6 +231,20 @@ model seasontickettype {
...
 
+/// price of ticket type for all events in a given season
+model seasontickettypepricedefault {
+  id                 Int                  @id @default(autoincrement())
+  seasonid_fk        Int
+  tickettypeid_fk    Int
+  price              Decimal              @default(0) @db.Money
+  concessionprice    Decimal              @default(0) @db.Money
+  season             seasons              @relation(fields: [seasonid_fk], references: [seasonid], onDelete: Cascade, onUpdate: Cascade)
+  tickettype         tickettype           @relation(fields: [tickettypeid_fk], references: [tickettypeid], onDelete: Cascade, onUpdate: Cascade)
```
henrykaus, line 242: Why does `Cascade` do for onDelete?

server/src/controllers/seasonTicketTypePriceDefaultController.ts
```diff
@@ -0,0 +1,183 @@
...
+    tickettypeid_fk: number;
+    price: number;
+    concessionprice: number;
+}
+
+/**
+ * @swagger
+ * /2/season-ticket-type-price-default/{id}:
+ *   get:
+ *     summary: get season ticket type prices for a given season
```
henrykaus, line 25: Might be good to say that is is by season id

server/src/controllers/seasonTicketTypePriceDefaultController.ts
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * @swagger
+ * /2/season-ticket-type-price-default/{id}:
+ *   put:
+ *     summary: update season ticket type prices for a given season
+ *     tags:
+ *     - Season Ticket Type Price Default API
+ *     parameters:
+ *     - $ref: '#/components/parameters/id'
+ *     security:
```
henrykaus, line 78: How do you specify whether you are updating, adding or deleting?

benfaustin95: The season ticket type prices are updated to match the array sent with the request body.

- If the ticket type exists in the table for the season and is not in the array it is deleted.
- If the ticket type is in the array but doesn't exist in the table for the given season it is created.
- If the ticket type both exists in the table for the given season and array it is updated.

server/prisma/schema.prisma
benfaustin95: If the ticket type being referenced is deleted the season ticket type price can not exist because there is no longer a ticket type to modify the price/concession price for. Similarly, if the season referenced is deleted there is no longer a season to attach the modified prices to.

server/src/controllers/seasonTicketTypePriceDefaultController.ts
henrykaus: Ok, that makes sense, do you think you could add the three cases to the swagger docs above. It looks like its possible to describe multiple request bodies.
image

henrykaus: The image is from https://swagger.io/docs/specification/describing-request-body/

server/prisma/schema.prisma
henrykaus: Ok perfect!

server/src/controllers/seasonTicketTypePriceDefaultController.ts
benfaustin95: Added examples

server/src/controllers/seasonTicketTypePriceDefaultController.ts
benfaustin95: Updated

henrykaus (approved):
Thanks for making the changes!

PR #479, opened 2023-11-21
TIC-139 Fixed Order Time

jonathantnguyen:
### Description
Fixed an issue where it would show the order time 8 hours ahead of the current time.
IE. a purchase at 1:00PM would show a time purchase at 9:00PM
Previous:
image
New:
image
*20:13 -> 8:13

When viewing purchase history within the contacts page, the time is correct as well.

### Risks
Unsure if being in a different timezone  will incorrectly set the time.

### Validation
Manual Testing

### Issue
[*Link to corresponding issue*](https://pdx-hkaus.atlassian.net/browse/TIC-139)

### Operating System
Win10/Chrome


henrykaus (approved):
Lets see if it works in dev

PR #486, opened 2023-11-22
Tic 127: Corrected issue to allow user to select new date after hitting "Choose Different Date" on Events Page

txiong002:
### Description
When choosing a showing to purchase. If you were to choose a date, and then choose a time. And then select a different date, the time would already be selected and you wouldn’t be able to purchase any tickets until going back to the main page or refreshing the page

### Risks
N/A

### Validation
Able to go through whole process, after changing to new date and time

### Issue
TIC-127

### Operating System
Window


henrykaus (approved):
Thanks for looking into this!

PR #489, opened 2023-11-22
FW23-15 Implemented Season Testing

ReidLuttrell:
### Description
Implemented season testing for navigate to seasons, add season, add event to season, edit season, remove event from season, remove season

### Risks
Edit button and delete button selector could be ambiguous if more Tooltip buttons are added later

### Validation
Tests passing

### Issue
https://trello.com/c/elsm5hZJ/25-implement-season-testing

### Operating System
Windows 11

### DIFF
--- a/client/e2e/pages/seasonsPage.ts
+++ b/client/e2e/pages/seasonsPage.ts
@@ -78,7 +78,7 @@ export class SeasonsPage {
 
   async addEventToSeason(event: EventsInfo) 
   {
-    const seasonLocator = this.page.getByTestId('season-event-section').filter({hasText: event.eventName});
+    const seasonLocator = this.page.getByTestId('season-event-card').filter({hasText: event.eventName});
     await this.addEvent.click();
     seasonLocator.getByTestId('event-add-to-season').click();
     await this.page.getByRole('button', { name: 'Continue' }).click();
@@ -104,7 +104,7 @@ export class SeasonsPage {
   {
     await this.page.locator(':text("' + season.seasonName + '")').click();
 
-    const seasonLocator = this.page.getByTestId('season-event-section').filter({hasText: event.eventName});
+    const seasonLocator = this.page.getByTestId('season-event-card').filter({hasText: event.eventName});
     seasonLocator.getByTestId('event-remove').click();
 
     await this.page.getByRole('button', { name: 'Continue' }).click();
--- a/client/e2e/pages/seasonsPage.ts
+++ b/client/e2e/pages/seasonsPage.ts
@@ -1,0 +1,165 @@
+import {type Locator, type Page, expect} from '@playwright/test';
+import {EventsPage} from '../pages/EventsPage';
+import { EventsInfo2 } from '../testData/ConstsPackage';
+
+export class SeasonsPage {
+  readonly page: Page;
+
+  // Homepage
+  readonly emailButton: Locator;
+  readonly manageTicketingButton: Locator;
+  readonly dashboardSeasonsButton: Locator;
+
+  // ../ticketing
+  readonly SeasonHeading: Locator;
+  readonly addSeasonButton: Locator;
+
+  readonly pageHeader: Locator;
+  readonly seasonNameBlank: Locator;
+  readonly seasonStartDate: Locator;
+  readonly seasonEndDate: Locator;
+  readonly imageURL: Locator;
+
+  readonly seasonOption1: Locator;
+  readonly seasonOption2: Locator;
+
+  readonly seasonEdit: Locator;
+  readonly seasonSave: Locator;
+  readonly seasonDelete: Locator;
+
+  readonly seasonContinue: Locator;
+
+  readonly DashboardButton: Locator;
+  
+  readonly addEvent: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+
+    this.emailButton= page.getByText(process.env.TEST_EMAIL as string);
+
+    this.manageTicketingButton=page.getByText('Manage Ticketing').first();
+    this.dashboardSeasonsButton=page.getByRole('button', {name: 'Seasons'});
+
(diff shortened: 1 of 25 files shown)


client/e2e/pages/seasonsPage.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  }
+
+  async addNewSeason()
+  {
+    await this.addSeasonButton.click();
+    await this.seasonNameBlank.fill('TestName');
+    await this.seasonStartDate.fill('2020-02-02');
+    await this.seasonEndDate.fill('2020-03-02');
+    await this.imageURL.fill('https://www.hindustantimes.com/ht-img/img/2023/08/25/550x309/international_dog_day_1692974397743_1692974414085.jpg');
```
snorq2, line 72: I suggest building an interface in ConstPackages to contain the season data.  This will enable a future improvement to create a version with unique identifiers and reduce the risk of flaky/failed tests interfering with other tests.

Also, set up the page object functions to accept arguments for the various season/ticket information instead of hardcoding them.  Then pass that information from the test file.  This increases your flexibility by allowing page functions to be reused in other tests.

client/e2e/pages/seasonsPage.ts
```diff
@@ -0,0 +1,165 @@
...
+    await seasonsPage1.goto();
+
+    // remove from default season
+
+    // click the first season that was created previously -- assuming its there
+    await this.page.locator(':has-text("Start Date")').last().click();
+
+    await this.page.getByRole('button', { name: 'Remove Event'}).last().click();
+
+    await this.page.getByRole('button', { name: 'Continue'}).click();
```
snorq2, line 100: I haven't checked the template code, but I would recommend checking to see if an id exists you can use.  If not, you can use a data-testid tag instead.

To differentiate between tags when multiple appear on a page, you can use a filter.  For example, if you added test ID's to the season card template and its relevant buttons, you could do this in your page object:

`const seasonLocator = this.page.getByTestId('season-card').filter({hasText: 'event name');
seasonLocator.getByTestId('delete-button').click();`

The first line first finds all the locators with the specified test ID, and then filters down to the one with the event name.  As long as the name is unique, this will work.  The second line then finds the locator with the test ID _inside_ of the first locator you found, and then clicks it.

client/e2e/pages/seasonsPage.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    // go back to seasons
+    const seasonsPage2 = new SeasonsPage(this.page);
+    await seasonsPage2.goto();
+
+    await this.page.locator(':text("TestName")').first().click();
+
+    await this.addEvent.click();
+    //await this.page.getByRole('button', { name: 'Add Event'}).click();
+    await this.page.getByRole('button', { name: 'Add to Season'}).first().click();
```
snorq2, line 110: I'd look to see if you can do something similar to lines 96-100 - use some kind of ID and filter to find what you need.  Anywhere you have to use a `.first` to find dynamic content (ie, an event or season) I'd highly recommend trying to find a more specific method to identify the locator.

client/e2e/pages/seasonsPage.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  async addEventToSeason() 
+  {
+    // notes:
+    // some goto methods are using loopholes to get to the page instead of the navigators
+
+    // create new event
+    const eventsPage = new EventsPage(this.page);
+    await eventsPage.goto();
+    await eventsPage.addnewevent(EventsInfo2);
```
snorq2, line 87: This should be in the test file, as part of your test setup.  Also, as a rule of thumb any added events should also be removed - it's not a huge issue in local, but up in dev the database is persistent.

client/e2e/seasonsPage.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    const seasonsPage = new SeasonsPage(page);
+    await seasonsPage.goto();
+  });
+
+test('addNewSeasonWithEvent',async({page})=>{
+    const seasonsPage = new SeasonsPage(page);
+    await seasonsPage.goto();
+    await seasonsPage.addNewSeason();
+    await seasonsPage.addEventToSeason();
+});
```
snorq2, line 14: As a rule of thumb, tests should be atomic - that is, don't rely on any pre-existing state at the start of your test, and clean up anything you add at the end.  I see there's some cleanup in future tests from the event added in here, but ultimately each test should contain its own cleanup.

A pattern we've found that's working so far is to put the test inside of a try block, and then putting all the cleanup inside of a finally block.  That way even if the test fails, the database is still cleaned up.  In this case, if possible I would suggest separating the logic that adds the seasons from the expects that are checking things, do all your add logic (both event and season adds) before the try block, all your expects inside of the try block, and then cleanup in the finally.  That said, this pattern won't necessarily work in all cases (say, you need to do some checks between adding different things).

snorq2 (commented):
Shared some practices we've discovered over the last couple of months messing with Playwright - let me know if you have any questions.  The `mainPage.ts` and `mainTest.spec.ts` aren't as tidy as I'd like, but they'll generally show examples of what I'm talking about.

client/e2e/seasonsPage.spec.ts
ReidLuttrell: Made the changes suggested and I believe I have covered all the points you brought up. Let me know if this is more what you had in mind!

jtng3:
Interesting, Reid I did not know you were working on testing. I was only aware Nate would be joining. I believe @Shihao126 had claimed seasons testing in the test bank and was working on these tests. Have you accessed the test bank yet?

jtng3 (approved):

jtng3:
tests only failing for ticket purchase (the date issue) for stripe in firefox for github actions. I am approving merge to main as sawyers branch will fix that issue.

PR #493, opened 2023-11-22
Add auto client unit test

ChentaoMa:
### Description
Added the client unit test

### Risks
null

### Validation
local

### Operating System
Windows

### DIFF
--- a/.github/workflows/client_test.yml
+++ b/.github/workflows/client_test.yml
@@ -25,9 +25,3 @@ jobs:
         run: npm run  test
         working-directory: ./client
       
-      - uses: actions/upload-artifact@v3
-        if: always()
-        with:
-          name: test-results
-          path: ./client/test-results
-          retention-days: 30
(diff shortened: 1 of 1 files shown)


.github/workflows/client_test.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          mkcert -install
+          mkcert localhost
+        working-directory: ./server
+        
+      - name: docker setup
+        run: docker-compose up -d
+
+      - name: 90 second wait for client and server start
+        run: sleep 90s
```
henrykaus, line 66: I don't actually know if these two steps are necessary as I'm able to run unit tests locally without having the environment up. This could mean that more than just these two steps could be removed from the pipeline

henrykaus (commented):
This is just a question but do we need to upload the playwright artifacts?

henrykaus (approved):
Thanks for slimming this down and making all the changes!

PR #494, opened 2023-11-22
[DEP-88] admin purchase e2 e tests

ncallon503:
### Description
Tests added for opening admin and ticket page, and purchasing a ticket for a customer as admin in both cases of recognized and new accounts

### Risks
I set wait functions in some parts of the code because in WonderTix it might search for a locator before the next pages finishes loading, not sure if this is the best way to solve this problem. Checking that the test worked is confirmed by "expect(page.getByText('Thank you for your purchase!') != null)" (reaching the final page) and also no errors being thrown, so not sure if there is a more efficient way of doing this as well.

### Validation
I ran the tests in all 3 browsers in playwright and they all passed, with both scenarios of unrecognized accounts and returning accounts that had the Stripe confirmation code pop-up.

### Issue
https://wondertix.atlassian.net/browse/DEP-88

### Operating System
Windows 11


server/prisma/legacy-data/events.yaml
```diff
@@ -0,0 +1,252 @@
+- eventid: 1
+  seasonid_fk: 1
+  eventname: Our Town
+  eventdescription: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+  active: false
+  seasonticketeligible: true
+  imageurl: https://www.arts.gov/sites/default/files/styles/portrait_crop/public/Our%20Town%20Cover%20new.jpg?h=52605a11&itok=PE1XH64E
+- eventid: 2
```
ben16williams, line 8: Is there any reason for saving older versions of `event.yaml` and `donations.yaml` in a new `legacy-data` folder? I think we could probably remove these and keep the ones we have updated in the `yaml-seeder-data` folder

henrykaus (commented):
@ncallon503 It looks like there are some merge conflicts on this one, is this PR still a work in progress?

henrykaus (commented):
@ben16williams This PR is ready for re-review.

client/e2e/pages/AdminPurchasePage.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  readonly eventDropdown: Locator;
+  readonly eventTimeDropdown: Locator;
+  readonly ticketTypeDropdown: Locator;
+  readonly checkoutButton: Locator;
+
+  readonly firstName: Locator;
+  readonly lastName: Locator;
+  readonly phone: Locator;
+  readonly email: Locator;
```
benfaustin95, line 21: It looks like the only issue is updating the page to handle the `Confirm Email` field.

benfaustin95 (commented):
It looks good! Just one small change that might need to be made.

client/e2e/pages/AdminPurchasePage.ts
henrykaus: Fixed

benfaustin95 (approved):

PR #495, opened 2023-11-25
[TIC-109] Hotfix to allow purchase of new showings

henrykaus:
### Description
New showings would be created with a time in the format of HHmmss+00 whereas seeded events were created with a time in the format HHmmss.SSS+00. This caused problems to some recent enhancements to how dates were handled in admin purchase. This was fixed.

image

Also allowed admins to purchase tickets from non-active events and sorting descending by eventid.

### Risks
None

### Validation
Tested with new and old events

### Operating System
Windows 11


0Tendo (approved):
good call! looks good.

PR #501, opened 2023-11-27
Fix event image in seasons page

HaliTran:
### Description
Fix error where default season image was being shown for an event

### Risks
Little to none

### Validation
Checked in the UI

### Issue
No issue

### Operating System
M2 Macbook


henrykaus (approved):
Looks good. Thanks for the quick fix. I’d wait for tests to pass before merging

PR #504, opened 2023-11-28
DEP-94: optimize deployment pipeline

jpb7:
### Description
- Upgrades builder to 8 CPUs (up from 2) for faster processing and parallel execution
- Runs E2E tests in parallel (in separate containers)
- Uses Kaniko instead of Docker for image layer caching
- Uses GCS bucket to cache `node_modules` across builds
- Fixes typo and removes problematic YAML from recent changes to Prisma

After adding E2E tests with long timeouts and Prisma seeders that took over 10 minutes to complete, we started hitting our 60-minute limit on Cloud Build. This branch reduces our build time on GCP back down to about 15 or 16 minutes.

### Risks
- Removing `PPH_transactions.yaml` is not desirable, but it's necessary for now so that the server can deploy to Cloud Run within the 15-minute time limit.
- Introduces a custom Kaniko image to enable use of Bash scripts as an entrypoint. Not much risk here; just some added complexity.

### Validation
- Successfully ran on GCP:
  - https://console.cloud.google.com/cloud-build/builds;region=us-west2/39bfaa17-fc66-401e-b3c0-f5e5b0fa91f0?project=wondertix-app

### Issue
- [*DEP-94: run E2E tests in parallel*](https://wondertix.atlassian.net/browse/DEP-94)

### Operating System
- Ubuntu 22.04


server/prisma/src/seedPPH_Opportunity.ts
jpb7, line 1: This typo was preventing the database from seeding properly.

client/e2e/pages/mainPage.ts
jpb7, line 1: This was throwing a type error without the explicit declaration.

client/e2e/EventsPage.spec.ts
jpb7, line 1: Forgot about this: 3 seconds is probably a bit harsh. Should probably be `30000` instead.

client/e2e/pages/mainPage.ts
```diff
@@ -122,7 +122,7 @@ export class MainPage {
   // String shift removes the first item in the array
   // Select a random option from the array, set that option, and return the text
   private async selectRandomOption(optionBox: Locator) {
-    const allOptions = (await optionBox.evaluate((sel) => {
+    const allOptions = (await optionBox.evaluate((sel: HTMLSelectElement) => {
```
jpb7, line 125: This was throwing a type error without the explicit declaration.

server/prisma/src/seedPPH_Opportunity.ts
```diff
@@ -17,7 +17,7 @@ async function seedPPHOpportunity(prisma: PrismaClient) {
       return;
     }
 
-    const yamlData = fs.readFileSync('./prisma/yaml-seeder-data/PPH_Opportunity.yaml', 'utf8');
+    const yamlData = fs.readFileSync('./prisma/yaml-seeder-data/PPH_opportunity.yaml', 'utf8');
```
jpb7, line 20: This typo was preventing the database from seeding properly.

dstad418: Good catch!

dstad418 (approved):
lgtm! Good catch on that typo, it never threw errors in my local environment, but I could see how cloud deployment could gum it up.

PR #514, opened 2023-12-01
Shrunk seeder files, made fixes team leads requested

dstad418:
### Description
Cut down seeder files to around 300 instances of data.
Reset the DB to test these changes, with no errors noted in docker logs or relational breaks
Changed output to only show in **_'prd'_**

### Risks
May cause issues if user doesn't re-migrate, and composes the changes with previous DB images. 
Potential relational breaks, or errors as noted in cloud deployment

### Validation
Brought the DB down and up again, no issues noted.

### Operating System
MacOS, M2 ARM


server/prisma/migrations/20231201205517_updated_pph_tables/migration.sql
henrykaus, line 1: Do you know why this migration was created?

dstad418: Not certain, I pulled from main before I made the changes, but that was a migration that was required for me to build the DB.
Looks like it's to help with integrity for the 'eventtickets' model, and to make the fk adjustments more robust.

henrykaus: Interesting, well it lines up with the current postgres schema so I don't see a problem

benfaustin95: I don't think that it actually changes anything so it should be good. 👍

henrykaus (approved):

PR #520, opened 2023-12-05
DEP-107 Rearranged testData structure

snorq2:
Refactored test data into classes
Refactored impacted tests

### Description
testData structure has been simplified.  All test interfaces, classes, and data are now contained in standalone files in the testData folder.  Common support functions are in testData/supportFunctions - these are used only by the test data objects, and are not intended to be used inside of either tests or page objects.

All test data objects have been refactored into interfaces and classes.  The factory functions have been removed.  Class and interface objects are now interchangeable - a new class object using an interface object must be called to take advantage of randomization.

##### CreditCard and ShowingInfo

These are simple interfaces with predefined consts available in the package.  Randomization is not appropriate for these data types, so the consts can be used directly.

##### CustomerInfo

Must call a new CustomerInfo object using a const to take advantage of randomization.  The data members will be populated in the class slightly differently depending on the value of the const data members:
fullName:  The value in the const object will be ignored, as the first name will always have a random value appended.
phoneNumber:  If a value exists in the const object, it will be used.  Otherwise a phone number will be randomly generated.
email:  If a value exists in the const object, it will be used.  Otherwise the first and last name, affixed with '@wondertix.com' will be used to automatically generate this value.

##### EventInfo

Must call a new EventInfo object using a const to take advantage of randomization.  eventFullName will be automatically generated from the other fields - the value in the const object will be ignored.

##### SeasonInfo

Must call a new EventInfo object using a const to take advantage of randomization.  seasonName will have a UUID affixed, but all other members will be taken from the const object.

### Risks
Limited backward compatibility - all tests not in main will need to refactor to use the test data objects properly.

The class and interface objects are interchangeable.  No errors will be raised if somebody uses a const instead of creating a class object.

### Validation
Tested in local.

### Issue
https://wondertix.atlassian.net/browse/DEP-107?atlOrigin=eyJpIjoiMmY3NmQwNTQ1NjJkNDRmYWE0OWQwZmQwNjA1M2Y1MDIiLCJwIjoiaiJ9

### Operating System
Windows 11


snorq2:
eventFullName has been removed.  The logic to build the appropriate string was moved into the EventsPage object and is no longer required in testData.

Shihao126 (approved):
You made it more modular, nice work Sawyer!

Issue #521, opened 2023-12-05
Wipe data from staging and production database tables


ben16williams:
Includes changing the deployment so that the staging and production tables aren't pre-populate with test data during deploy

PR #528, opened 2023-12-12
Conditionally seed database 

ben16williams:
### Description
Adds a `SHOULD_SEED` environment variable so that the database is only seeded with essential data on startup if it's set to `true`. The essential data being two ticketTypes that are each expected to have specific id values in other places in code, so those two types have to be specifically seeded (the altered sequence for the id value can be found at the end of the first db migration), as well as all the `date` table values 

The goal of this is to have staging and production environments that aren't filled with test data every time code's promoted 

Created the secret in gcp with versions for each env [here](https://console.cloud.google.com/security/secret-manager/secret/SHOULD_SEED/versions?project=wondertix-app)

### Risks
May have missed some environment variable configuration in some place(s)

### Validation
DB is only seeded with expected values locally for both true and false values 

### Issue
https://github.com/WonderTix/WonderTix/issues/521

### Operating System
win10

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -19,6 +19,9 @@ CI=false
 # dates, ticket types, discounts etc. See seed.ts in server/prisma
 MINIMAL_SEED=false
 
+# Whether or not to seed the database with any test data
+SHOULD_SEED=false
+
 # For Playwright E2E Testing
 TEST_EMAIL=none
 TEST_PASSWORD=none
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -180,6 +180,7 @@ steps:
       - 'SERVER_CPU'
       - 'SERVER_MEMORY'
       - 'SERVICE_ACCOUNT'
+      - 'SHOULD_SEED'
       - 'SQL_INSTANCE'
 
   # Send traffic to new server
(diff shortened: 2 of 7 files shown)


ben16williams:
Would like to get this reviewed by @jpb7 or @jtng3 🙏 otherwise will merge it down in a day or so

jtng3 (commented):
lgtm. @jpb7 what do you think?

jpb7 (approved):
Looks fine to me.

ben16williams:
gcbrun

jpb7:
We need to update the readme: `gcbrun` is no longer used.

Based on the logs, it looks like the new `SHOULD_SEED` variable is not being passed to the `deploy-test-server` step of the top-level `cloudbuild.yaml`. `SHOULD_SEED` will need to be added to the `secretEnv` of each step in the `cloudbuild.yaml` where it's being used.

My bad: should've caught that in code review.

ben16williams:
> We need to update the readme: gcbrun is no longer used.
> 
> Based on the logs, it looks like the new SHOULD_SEED variable is not being passed to the deploy-test-server step of the top-level cloudbuild.yaml. SHOULD_SEED will need to be added to the secretEnv of each step in the cloudbuild.yaml where it's being used.
> 
> My bad: should've caught that in code review.

Ah no worries, I was wondering about that since I didn't see it on any other recent PRs. I just ran the trigger manually in GCP and that seemed to work. I'll update that `cloudbuild.yaml` for the env var and the `README`

Issue #540, opened 2023-12-21
Add ability to create a contact


ben16williams:
Ideally this would include viewing, editing, deleting, and adding a new contact but only having the ability to add a contact in the admin UI is fine for first pass

henrykaus:
Viewing and deleting should already be implemented, but I'll double check next week. And Ill create a ticket for added the ability to edit an existing contact.

PR #543, opened 2023-12-29
[WT-540] Create Contact

henrykaus:
### Description
* Created a `ContactPopUp` that allows users to fill in contact information. 
* On a successful creation, the user is brought to the "More Info" page for that contact.

### _SEE RISKS SECTION_

image
image
image

https://github.com/WonderTix/WonderTix/assets/92124260/a4c71777-9155-4fce-b2b8-45271166b310

https://github.com/WonderTix/WonderTix/assets/92124260/4303cbf0-a2ec-4069-9715-20274076f8b4

### Risks
* This will require reseeding the deployed environments if they have seed data in them as the new seed data doesn’t have duplicated email addresses; and the schema update does not allow for non-unique emails

### Validation
Manually tested

### Issue
#540 

### Operating System
macOS M2


server/prisma/migrations/20230524160200_init/migration.sql
```diff
@@ -9,7 +9,7 @@ CREATE TABLE "contacts" (
     "contactid" SERIAL NOT NULL,
     "firstname" VARCHAR(255),
     "lastname" VARCHAR(255),
-    "email" VARCHAR(100),
```
henrykaus, line 12: Ideally, this should be in a new migration file

ben16williams:
> This will require reseeding the deployed environments if they have seed data in them as the new seed data doesn’t have duplicated email addresses; and the schema update does not allow for non-unique emails

This constraint at the database level makes sense to me. I don't _think_ having existing contact records with duplicate emails should cause any problems once this constraint is applied. It doesn't look like there was that many of them so we could manually delete them instead of trying to reseed dev entirely, which I think would involve either deleting everything from all tables, deleting everything from the contacts table, or deleting the db entirely. I'm not sure exactly which one of those options would actually cause the table(s) to get reseeded though

ben16williams (approved):
Looks great!

One non-blocking question - how did you generate the migration? I've been trying to get it to generate using `npx prisma migrate dev --create-only ` after updating the  `schema.prisma` with the `DATABASE_URL` pointed to a locally running postgres container, but haven't been able to get it to work. Or did you create the migration manually?

henrykaus:
> Looks great!
> 
> One non-blocking question - how did you generate the migration? I've been trying to get it to generate using `npx prisma migrate dev --create-only ` after updating the `schema.prisma` with the `DATABASE_URL` pointed to a locally running postgres container, but haven't been able to get it to work. Or did you create the migration manually?

You have to do a couple steps:
1) Copy your `.env` into the server directory.
2) Add the following variable to that `.env`:
```
DATABASE_URL=postgresql://${PG_USER}:${PG_PASS}@${PG_HOST}:${PG_PORT}/${PG_DB}?schema=public
```
3) Remove the `CI` variable from that file (I don't know why but Prisma specifically throws a fit when its included)
4) Have wondertix running locally via docker
5) Go to `WonderTix/server` and create the migration with `npx prisma migrate dev --name name_here` where `name_here` is the name you want the migration to be called, ex. `unique_contact_emails`

ben16williams:
> > Looks great!
> > One non-blocking question - how did you generate the migration? I've been trying to get it to generate using `npx prisma migrate dev --create-only ` after updating the `schema.prisma` with the `DATABASE_URL` pointed to a locally running postgres container, but haven't been able to get it to work. Or did you create the migration manually?
> 
> You have to do a couple steps:
> 
> 1. Copy your `.env` into the server directory.
> 2. Add the following variable to that `.env`:
> 
> ```
> DATABASE_URL=postgresql://${PG_USER}:${PG_PASS}@${PG_HOST}:${PG_PORT}/${PG_DB}?schema=public
> ```
> 
> 3. Remove the `CI` variable from that file (I don't know why but Prisma specifically throws a fit when its included)
> 4. Have wondertix running locally via docker
> 5. Go to `WonderTix/server` and create the migration with `npx prisma migrate dev --name name_here` where `name_here` is the name you want the migration to be called, ex. `unique_contact_emails`

Ah ty! I had some weirdness going on with a locally running postgres server but was eventually to get prisma pointed at the right db using this

PR #545, opened 2023-12-31
Add missing comma between env vars in server deploy script

ben16williams:
### Description
Add missing comma between env vars in server deploy script

### Risks
none

### Validation
can only exercise the updated script after the change is merged 

### Operating System
win10


henrykaus (approved):
Looks good, not sure what it fixes, but it looks like something was wrong

ben16williams:
> Looks good, not sure what it fixes, but it looks like something was wrong

This was causing the `SHOULD_SEED` environment variable was being including as part of the `ROOT_URL`. eg the value would be `ROOT_URL=https://wtix-server-dev-3ps4rkx4ga-uw.a.run.app SHOULD_SEED=true` instead of two distinct environment variables. Conveniently, you can see all the environment variables a cloud run revision has configured from the service details in gcp ([stg for instance](https://console.cloud.google.com/run/detail/us-west1/wtix-server-dev/revisions?project=wondertix-app))

PR #547, opened 2024-01-02
WT-521 Reset ID sequences

ben16williams:
### Description
In preparation for a production instance, this adds a database migration that resets most tables primary key `id` sequence to one, except for the `tickettype` table that has two ticket types that reserve the id values of `0` and `1`. 

This also updates the seeding script and data so that line items id values are auto-incremented according the `SERIAL` pseudo-type so that, post-seeding, new lines items created in a table do not collide with existing id values like they would've previously. It appears that postgres is unable to track id values when they are set explicitly with the old seeding scripts. Technically did not have to remove the hard-coded `id` values from the yaml data, but did so that they aren't used as a reference and to avoid any confusion. 

### Risks
There are a couple of instances where certain table entries are expected to have a certain id value, like with the two `tickettypes `in the `tickettype` table. The only other place I found this to be the case was with `events` where some front-end components look like they use the event with a `0` id as a template for creating new events. Was able to get around this by having the `events` table start incrementing from `1`, like most other tables. 

### Validation
Applied the migration locally and ran the seeding scripts with no issues. 

### Issue
#521

### Operating System
win10


henrykaus (approved):
This makes sense to me!

PR #550, opened 2024-01-02
[WT 530] Updated Seasons so ticket type information persists on season creation and added season information validation

### LINKED ISSUE #512: Do not allow Season end date before start date
In the UI of the Seasons page, you are allowed to set the enddate to anytime regardless of the start date. It would make sense to require the end date to be anything greater than or equal to the start date. I believe that `<input type='date' ... />` has a `min` prop that could be used.

### LINKED ISSUE #530: Season ticket type information doesn't persist when creating a season
When creating a new season, the information filled into the ticket type table is ignored. The ticket type data needs to be handled appropriately and added to the season once as it is created. Note that once the season exists, then updates to the table work just fine.

**Creating a new season with a default ticket type price set:**

image

**After creating by clicking Save:**

image

benfaustin95:
### Description
The season creation process has been updated so that season ticket types persist when the season is initially created as well as when the season is edited. Additionally I added season information validation. 

### Risks
A bug went unnoticed

### Validation
Manual testing of the processes. 

### Issue
closes #530
closes #512 

### Operating System
*If you're on Mac, please include whether you are on an Intel or M-series CPU*


ben16williams (approved):
Looks good 👍

PR #551, opened 2024-01-04
[WT-544] Add ability to edit a contact

henrykaus:
### Description
Added ability to edit a contact, and refactored the deletion of a contact.
* Added new buttons for editing and deleting a contact
* Added confirmation popup when deleting a contact
* When deleting a contact, the page now updates on deletion so the contact disappears (instead of persisting until a refresh) [this is also the case for the new editing]
* If a contact can not be deleting, a PopUp shows the error
* General visual and accessibility improvements
* Added "No results" message when a search returns nothing

Just to note:`ContactResults` got refactored into `ContactCard`.

image
image
image
image

### Risks
I deleted a reference to `ContactResults` within a Task-based Search component. But after looking into things, it appears that this Task component (as well as many others) were replaced with the recent task management overall. TLDR; these components aren't used anywhere anymore.

### Validation
Manual testing with various scenarios.

### Issue
#544 

### Operating System
macOS M2

### DIFF
--- a/client/src/components/CRM/Contacts/ContactCard.tsx
+++ b/client/src/components/CRM/Contacts/ContactCard.tsx
@@ -1,42 +1,44 @@
 import React from 'react';
 import {useState} from 'react';
-import {Tooltip} from '@mui/material';
 import {useNavigate} from 'react-router-dom';
+import {Tooltip} from '@mui/material';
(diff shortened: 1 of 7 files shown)


ben16williams (approved):
lgtm, that error handling looks great

PR #559, opened 2024-01-07
Revert "Minimal change in the server directory to get migrations to run on deploy (#556)"

ben16williams:
### Description
Reverts "Minimal change in the server directory to get migrations to run on deploy (#556)"

Now need some change to the server to trigger the staging database migrations when I deploy to it

### Operating System
win10


ben16williams:
Force merging because this has no risk and is blocking production deployment

PR #561, opened 2024-01-08
Minimal change to the client to get a new revision to build in cloud run

ben16williams:
### Description
Minimal change to the client to get a new revision to build in cloud run

### Operating System
win10


ben16williams:
Force merging this since there is no risk and I need a client revision to build with config changes in dev. Will write a ticket for looking into a better way of doing this

Issue #516, opened 2023-12-02
Contacts page missing information


henrykaus:
When looking at a contact, there are several fields that do not show information. There are also many UI issues (especially on a mobile-screen size) with lacks of spacing.

This is correct (although the Donorbadge field is blank):
![image](https://github.com/WonderTix/WonderTix/assets/92124260/2b799d16-0dde-4fea-bbee-fc8085de1561)

After clicking `Show All Information`, the address is missing:
![image](https://github.com/WonderTix/WonderTix/assets/92124260/736a8161-a1ce-412b-b93b-6418b9f03ed7)

The spacing on the side of the screen should mimic other areas of the admin panels (i.e. reduce the left-right margins):
![image](https://github.com/WonderTix/WonderTix/assets/92124260/4d4b3243-800b-4232-8e65-f1ac431c2ecb)

The order date and times should be formatted in a readable manner and there should be a gap between the order cards:
![image](https://github.com/WonderTix/WonderTix/assets/92124260/14402d20-a93d-4abb-922a-e9da75b7ef08)

Also, if a contact doesn't have any orders, the message below should be a little less cryptic, something like `<person> hasn't made any orders.`.
![image](https://github.com/WonderTix/WonderTix/assets/92124260/f7ce85d2-d18d-4ca6-a2c9-0d904b932500)

The Refund button is also non-functional. Although, this should be discussed with the sponsor if they should be able to refund right from here, rather than just from the working Refund page.

ben16williams:
This is a high priority. The refund page is sufficient for now and we can remove the refund button from this order view.

ben16williams:
Showing all information for a contact should also show donations that they've made and the order information should be expanded to include all the items included in the order (showings, events, seasons, etc)

PR #562, opened 2024-01-10
[WT-516] Added order/donation history for a contact

henrykaus:
### Description
Improved the Single Contact Page to include order and donation history for that contact.
* Just to note that in this PR, there is no visual indication that a donation is associated with a ticket order.

image

image

### Risks
Haven't been able to test real donations since Im having issues with the Stripe webhook, but donations are making their way to the DB in staging, so I imagine that won't be an issue.

### Validation
Manual Testing.

### Issue
#516 

### Operating System
macOS M2


ben16williams:
I went through and cleaned up a lot of the Auth0 stuff over the weekend and yesterday and I meant to review the automated tests. I'll bet there is some cAuth0 client configuration that is mismatched. Taking a look now

client/src/components/CRM/Contacts/ContactOneResult.tsx
```diff
@@ -1,272 +1,271 @@
...
+      </header>
+      <div className='grid md:grid-cols-2'>
+        <article>
+          <p className='flex flex-row gap-3 text-lg w-full'>
+            <span className='font-semibold'>Order Date:</span>
+            <span>{format(date, 'MMM dd, yyyy')}</span>
+          </p>
+          <p className='flex flex-row gap-3 text-lg mt-1 w-full'>
+            <span className='font-semibold'>Order Time:</span>
+            <span>{format(date, 'h:mm a')}</span>
```
ben16williams, line 174: We should include the timezone here and with the `eventDate` in the order item

server/src/controllers/contactController.ts
```diff
@@ -243,6 +243,176 @@ contactController.get('/:id', async (req: Request, res: Response) => {
...
+                          },
+                        },
+                        eventinstances: {
+                          select: {
+                            eventdate: true,
+                            eventtime: true,
+                            events: {
+                              select: {
+                                eventname: true,
+                              },
```
ben16williams, line 313: It would be great it the `orderitems` view included the season that the event is in, if any. That might be too much of a lift for this PR though? It's not a blocker for me if you want to move that into another ticket

henrykaus: This should be pretty easy

henrykaus: image
(event not in season and event in season)

client/src/components/CRM/Contacts/ContactOneResult.tsx
henrykaus: It looks like a timezone isn't stored in the database with the time stamp (it is just converted to UTC). And when the time is displayed to the user on the frontend, it is converted to their local timezone (so PST in most cases). Do we just want to display the user's local timezone?

ben16williams: Hm it does look like for either of these timestamps they don't get saved with the timezone.. it also doesn't look like they are not converted to local time. If I create an order at 2pm my time (PST) then it gets saved as `14:00:00+00:00` and displayed back as 2pm. Maybe I'm missing something about these timestamps, but I would expect it to get saved as `22:00:00-08:00` where `22:00:00` is the UTC time and `-08:00` is the timezone offset. Either way, I'm ok with just keeping an eye on it and worrying about it if it starts causing problems

ben16williams (approved):

client/src/components/CRM/Contacts/ContactOneResult.tsx
ben16williams: I got those backwards my bad, I would expect the timestamp to be `14:00:00-08:00` where `14:00:00` is the local time PST and `-08:00` is the offset from UTC time

henrykaus: Yeah, that's what we'd like, but it looks like were saving the time as a DateTime where it takes the form `1970-01-01T04:05:06` and then we're just extracting the time.

PR #565, opened 2024-01-12
[WT-542] Added email, phone, and address to Door List

henrykaus:
### Description
Updated doorlist API and door list front end to display contact email, phone number, and address.

Also updated the doorlist to export with a more informative title, rather than just "WonderTix".

image

**CSV File:**
image

### Risks
May have missed a spot when updating the API.

### Validation
Manual testing of doorlist.

### Issue
#542 

### Operating System
macOS M2


benfaustin95 (approved):

ben16williams:
My internet is currently down so I can't properly review this, but we'll also want to list the ticket type in the door list. That might be trickier since one contact can have multiple ticket types to the same show, but I'm not sure what the logic for fetching the door list looks like. It's lower priority so I'll write a ticket unless you wanted to include it here

Issue #566, opened 2024-01-12
Add "Available" column for showing ticket types


henrykaus:
Add a column in the event information page to display how many tickets for a ticket type are still available.

PR #567, opened 2024-01-12
[WT-566] Added "Sold" column to showing ticket types

henrykaus:
### Description
Added a `Sold` column that displays how many tickets of a particular ticket type in a particular showing have been sold.

image

### Risks
There could be an area where the new `ticketssold` attribute could cause issues if it is being passed around inappropriately.

### Validation
Manual testing (see screenshot).

### Issue
#566 

### Operating System
macOS M2

### DIFF
--- a/server/src/controllers/eventInstanceController.ts
+++ b/server/src/controllers/eventInstanceController.ts
@@ -350,6 +350,11 @@ eventInstanceController.get(
             ticketrestrictions: {
               include: {
                 tickettype: true,
+                eventtickets: {
+                  where: {
+                    singleticket_fk: {not: null},
+                  },
+                },
               },
             },
           },
@@ -362,6 +367,7 @@ eventInstanceController.get(
             price: restriction.price,
             concessionprice: restriction.concessionprice,
             ticketlimit: restriction.ticketlimit,
+            ticketssold: restriction.eventtickets.length,
             description: restriction.tickettype.description,
           })),
         })));
--- a/client/src/components/Ticketing/ticketingmanager/Event/components/EventShowingForm.tsx
+++ b/client/src/components/Ticketing/ticketingmanager/Event/components/EventShowingForm.tsx
@@ -32,7 +32,11 @@ export const EventShowingForm = (props: EventShowingFormProps) => {
     defaulttickettype: 1,
     purchaseuri: 'http://null.com',
     instanceTicketTypes: initialValues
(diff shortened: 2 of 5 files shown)


ben16williams:
Because PPH request seems a little vague here, I'm leaning towards having the new column be the number of tickets sold of each type as opposed to the number of tickets available

From the screenshot in the description, I understand the reason behind the default ticket type and why there are still 98 `Available Tickets` to the show despite the number of `Available` tickets left for the two ticket types combined suggesting that there should only be 96 `Available Tickets` left, but I could see be confusing for someone who doesn't know the logic

Having the new column be the number of each type sold should eliminate that confusion and each show would have an at-a-glance view of both how many tickets remain for the show and how many have been sold without having to do some quick maths.

server/src/controllers/eventInstanceController.ts
```diff
@@ -350,6 +350,11 @@ eventInstanceController.get(
             ticketrestrictions: {
               include: {
                 tickettype: true,
+                eventtickets: {
+                  where: {
+                    singleticket_fk: {not: null},
+                  },
+                },
```
ben16williams, line 357: It looks like tickets sold of any other type also creates an `General Admission` ticket record in the `eventtickets` table which seems to throw off the count of the sold default `General Admission` tickets. What's the relation between `eventtickets` and `singletickets`? I'm not sure what the purpose of having a separate `singletickets` table is?

ben16williams: Ah I see, all available tickets are added to the `eventtickets` table and a new `singleticket` is created whenever one of `eventtickets` is reserved. I'm still not sure why a `singleticket` needs to map to multiple `eventtickets`, that seems to just be throwing off the count of each tickettype sold

benfaustin95: That sounds good, the implementation was pretty similar to what we have now so not much needed to be changed. I made most of the changes and am just going through the backend to make sure I didn't miss anything that is effected. Then I'll do some testing and submit a PR.

henrykaus: Ok, sweet, thanks @benfaustin95, should this PR go through before or after your changes?

benfaustin95: I think you should be able to merge your PR before the changes, the sold count for general admission will just be off temporarily.

benfaustin95 (approved):

Issue #553, opened 2024-01-05
Database migration for new contacts fields


ben16williams:
Add new fields to the contacts table for city, state, postal code, country, and created date 

Need these fields to migrate PPH data and the address information should be more complete anyways

ben16williams:
blocked by https://github.com/WonderTix/WonderTix/pull/547

Issue #552, opened 2024-01-05
Use additional contact address information


ben16williams:
Update all server routes for managing contacts with new fields and update all UI elements with inputs for new fields (cart checkout, creating contact, modifying contact, contact information view, etc.)

PR #569, opened 2024-01-14
WT-553 & WT-552 Add and use new contacts address fields

ben16williams:
### Description
- Creates a new migration for adding fields for `city`, `state`,  `country`, `postalcode`, and a timestamp to `contacts` records
- Uses the new fields everywhere in the frontend including in the contact checkout form, contacts search cards, and admin edit contact card
- Adds a new field to the contact checkout page for confirming the email address
- In the contact checkout page, any comments that a user adds will be appended to the seating accommodation when creating the contact
- Updates contacts seeding script and yaml data with some test data for new fields 

image
image
image

### Risks
Contact information is used in quite a few places in the client so I may have missed something there

### Validation
Did a lot of manual testing locally and verified that the new fields are persisted when creating a contact by both purchasing a ticket and manually creating one as an admin. Also verified that the fields are saved when updating a contact. 

### Issue
https://github.com/WonderTix/WonderTix/issues/552 and https://github.com/WonderTix/WonderTix/issues/553

### Operating System
win10


ben16williams:
> This looks pretty good. I think you may have missed the AdminCompleteOrderForm as that also gathers contact information.

Ah nice catch

> And then just a visual thing (but not blocking), I think it might make some sense in the ContactPopUp to have the City and State inputs next to each other on the same line and then Country and Postal Code on the same line. This will also help the form be a bit more usable on lower-resolution screens since it uses up less vertical space.

That looks a lot better to me too https://github.com/WonderTix/WonderTix/pull/569/commits/e81d1d2da595aae926f49fd93763993aca60e415

henrykaus (commented):
Just a couple questions. 1) It doesn't look like the `visitSource` (How did you hear about us?) is hooked up, does that matter? 2) When going through admin purchase, I think the comments are ignored when the accommodation isn't set to `Other` (the old flow). 3) Should the createddate be displayed anywhere?

ben16williams:
> 1) It doesn't look like the visitSource (How did you hear about us?) is hooked up, does that matter?

Yea that should be wired up. Create a ticket https://github.com/orgs/WonderTix/projects/3/views/1?pane=issue&itemId=50484211


> 2) When going through admin purchase, I think the comments are ignored when the accommodation isn't set to Other (the old flow).

Nice catch, updated https://github.com/WonderTix/WonderTix/pull/569/commits/764389c4304a23bac611e9572d9ca1502ac0c7de


> 3) Should the createddate be displayed anywhere?

Yea, meant to add that https://github.com/WonderTix/WonderTix/pull/569/commits/97533553200ba9f2df79ca0168f12bda033f7a92

henrykaus (approved):
Looks good!

PR #571, opened 2024-01-15
Add more detailed instructions for using Stripe locally and remove Auth0 client_secret env

ben16williams:
### Description
Attempts to make the Stripe setup less ambiguous and the auth0 `client_secret` apparently is not needed or used for the front-end react app to authenticate with Auth0

### Risks
Possibly the `client_secret` has some use  in the authentication process that I'm missing

### Validation
Spun up locally and had no problems logging in and doing all the admin things 

### Operating System
win10


henrykaus (approved):
Instructions are clear and nice catch on the AUTH0 secret, should help with the App's security

Issue #574, opened 2024-01-16
Refunded order items do not show in a contacts order history


ben16williams:
The `Refunded` field is updated but the orderitem is gone

![Image](https://github.com/WonderTix/WonderTix/assets/55893544/eb91c40b-1739-44cc-a1aa-817397d5f4d4)

Issue #570, opened 2024-01-14
Include ticket type(s) in doorlist


ben16williams:
The doorlist needs to show the ticket type along with the contact information and there needs to be a way to represent a contact that has purchased multiple different ticket types to the same show in the doorlist. The ideal solution would be to list all the owned tickettypes with the amount of each in the same line with the contact information. Another possible solution might be to have a new line for each ticket type with the same contact information.

Issue #568, opened 2024-01-13
Add indicator that a donation is associated with an order


henrykaus:
In the Contact information screen, a donation associated with an order should appear as a part of that order rather than in the separate tab.

Also, if more than one of the exact same ticket is purchased, then the quantity should be accurately depicted rather than repeated ticket items.

Issue #458, opened 2023-11-15
If amount is too small, stripe will throw error


henrykaus:
If you were to purchase tickets via admin purchase that were greater than 0 but very small, like `$0.01`, then stripe will throw an `amount_too_small` error like the image. This is not caught by WonderTix so no error is displayed. We should likely have this error displayed in a `PopUp`.
![image](https://github.com/WonderTix/WonderTix/assets/92124260/4f194df7-6cb0-4322-ba64-0f3d935b9c15)

PR #581, opened 2024-01-19
[WT 580] Simplify Ticketing Process & Update Doorlist

benfaustin95:
### Description

- Simplified ticketing process  by moving towards a schema that generates tickets on order rather than when ticket restriction is created/edited.  
- Removed  seeding data that is no longer applicable and added seeding of random orders. 

image
image
image

### Risks
Possibly missed a feature effected by the change in order processing. 

### Validation
Manual testing of event page, order processing, and doorlist. 

### Issue
#580 
#574
#570 
#568
#458 
#370 

### Operating System
M1 Mac


client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
```diff
@@ -6,72 +6,71 @@ import {titleCase} from '../../../../utils/arrays';
...
- */
-const renderCheckin = (params: GridCellParams) => {
-  const isDisabled = params.getValue(params.id, 'lastname') === 'OPEN SEATS';
+interface RenderCheckinProps {
+  params: GridCellParams;
+}
+
+const RenderCheckin = (props: RenderCheckinProps) => {
+  const {params} = props;
+  const {token} = useFetchToken();
```
henrykaus, line 22: Instead of fetching a token for each Checkin, would it make more sense to fetch 1 token in the doorlist component, and then pass it in through the props?

henrykaus (commented):
Everything appears to work. The only thing I worry about is that it isn't super known that the `Total Tickets` is the overall cap, and that PPH would have to set up their showing properly to allow for the ability to sell all of those seats should one ticket type not sell completely. But maybe that's a non-issue. Might want @ben16williams's opinion

ben16williams:
> The only thing I worry about is that it isn't super known that the `Total Tickets` is the overall cap, and that PPH would have to set up their showing properly to allow for the ability to sell all of those seats should one ticket type not sell completely. But maybe that's a non-issue. Might want @ben16williams's opinion

It's a small learning curve for giving more flexibility over event/showing creation and not being bound to having to have the same one ticket type for every showing. In PPHs case, sometimes they have shows that are only pay what you can tickets, and they might not necessarily want to have to offer a general admission along side that. But now if they wanted to they would have either option

ben16williams:
We could remove the `defaulttickettype` from the `eventinstances` table and remove it from the `Manage Ticket Types` page or make it so that it is modifiable/deletable

ben16williams:
If `singletickets` are one to one with `eventtickets` since a `singleticket` isn't mapped to both a 'custom' and a default ticket type in the `eventinstace` table now, then I'm not sure if there is use for a separate `singleticket` table? There may be some use there with ticket exchanges, but you'd still have to update the `eventticket` table either way when doing an exchange, whether that be modifying the `singleticket_fk` or the `orderitemid_fk` if it's moved to the `eventticket` table. It would be nice to reduce complexity where we can but also might be too much of a bother at this point so not a blocker for me

client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
```diff
@@ -6,72 +6,71 @@ import {titleCase} from '../../../../utils/arrays';
...
+      <ul className={'w-full h-[80%] overflow-scroll'}>
+        {
+          typeof ticketTypes === 'object' &&
+            Object.keys(ticketTypes).map((key, index) => (
+                <li key={`${params.row.id}-${index}`}>{key} - {ticketTypes[key]}</li>
+            ))
+        }
+      </ul>
+  );
+};
```
ben16williams, line 73: This looks great, nice solution! I also did not realize how robust that search bar already is in that you can toggle which columns are displayed and much more, definitely we'll want the same kind of thing in the contacts search. A couple of things though - the blocking one is that when exporting to excel, the ticket type(s) show up as `[object Object]`, and the less important one is that it would be great if it was possible to resize each column. for instance, my email ben16williams@gmail.com is too long and gets cut off so that the only way I can see the whole value is by hovering over it with my mouse (I could be being dumb and missing how you're suppose to do this though)

client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
benfaustin95: That makes sense, I just have to move the columns definition into the scope of the doorlist component.  👍

benfaustin95:
> If `singletickets` are one to one with `eventtickets` since a `singleticket` isn't mapped to both a 'custom' and a default ticket type in the `eventinstace` table now, then I'm not sure if there is use for a separate `singleticket` table? There may be some use there with ticket exchanges, but you'd still have to update the `eventticket` table either way when doing an exchange, whether that be modifying the `singleticket_fk` or the `orderitemid_fk` if it's moved to the `eventticket` table. It would be nice to reduce complexity where we can but also might be too much of a bother at this point so not a blocker for me

Given that we have some more time now that the deadline is extended It might be a good idea to refactor the current schema a little bit. I  think we could reduce the complexity by removing the singleticket and eventtickets tables and moving towards generating tickets upon order. I worked on a possible schema that helps reduce some of the complexity while also addressing some of the existing issues (individual item refunds, subscriptions, losing order details on refund). I dropped a link to it in the contractor Slack if you guys want to take a look.

client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
ben16williams: Ooof of course it is. The name, email and ticket types are probably the most important text fields that should have enough space

client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
benfaustin95: Fixed

client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
benfaustin95: Fixed

client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
```diff
@@ -115,6 +83,38 @@ const DoorList = (): ReactElement => {
...
+        },
+        {field: 'email', headerName: 'Email', width: 200},
+        {field: 'phone', headerName: 'Phone Number', width: 130},
+        {field: 'vip', headerName: 'VIP', width: 65, renderCell: renderCheckbox},
+        {
+            field: 'donorBadge',
+            headerName: 'Donor',
+            width: 65,
+            renderCell: renderCheckbox,
+        },
```
henrykaus, line 114: For both of these, I think it would probably make sense to just make them have a `type: boolean`, rather than a `renderCell: renderCheckbox` since it doesn't need to be an input, just a display

server/prisma/migrations/20240131062714_update_order_process/migration.sql
```diff
@@ -0,0 +1,212 @@
+/*
```
henrykaus, line 1: Just a note but seeing as migrations are really useful for environments where we care about keeping the contents of the database, it might make sense to at some point, delete all the previous migrations, and start fresh on the migrations seeing as they are building up and don't add much value at the moment.

server/prisma/migrations/20240131062714_update_order_process/migration.sql
```diff
@@ -0,0 +1,212 @@
...
+ADD COLUMN     "orderid_fk" INTEGER NOT NULL,
+ALTER COLUMN "amount" SET NOT NULL,
+ALTER COLUMN "frequency" SET NOT NULL;
+
+-- AlterTable
+ALTER TABLE "eventinstances" DROP COLUMN "defaulttickettype";
+
+-- AlterTable
+ALTER TABLE "events" ALTER COLUMN "eventdescription" SET NOT NULL,
+ALTER COLUMN "eventdescription" SET DEFAULT '',
```
henrykaus, line 92: Any reason why `eventdescription` needs to be set if it is defaulting to an empty string?

server/prisma/migrations/20240131062714_update_order_process/migration.sql
```diff
@@ -0,0 +1,212 @@
...
+ALTER COLUMN "active" SET NOT NULL,
+ALTER COLUMN "active" SET DEFAULT true,
```
henrykaus, line 94: I imagine that events should be `inactive` by default so they don't appear on the customer site right off the bat.

server/prisma/migrations/20240131062714_update_order_process/migration.sql
```diff
@@ -0,0 +1,212 @@
...
+ALTER COLUMN "seasonticketeligible" SET NOT NULL,
+ALTER COLUMN "seasonticketeligible" SET DEFAULT true;
```
henrykaus, line 96: Is there a region on the frontend where this value can be changed?

server/prisma/migrations/20240131062714_update_order_process/migration.sql
```diff
@@ -0,0 +1,212 @@
...
+
+-- AlterTable
+ALTER TABLE "orders" DROP COLUMN "orderdate",
+DROP COLUMN "ordertime",
+DROP COLUMN "ordertotal",
+DROP COLUMN "refund_intent",
+ADD COLUMN     "orderdateandtime" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
henrykaus, line 103: If its not too much effort, might be worth to just change the name to `orderdatetime` for simplicity

server/prisma/migrations/20240131062714_update_order_process/migration.sql
```diff
@@ -0,0 +1,212 @@
...
+
+    CONSTRAINT "order_ticketitems_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "ticketitems" (
+    "id" SERIAL NOT NULL,
+    "order_ticketitemid_fk" INTEGER NOT NULL,
+    "ticketrestrictionid_fk" INTEGER NOT NULL,
+    "donated" BOOLEAN NOT NULL DEFAULT false,
```
henrykaus, line 140: Do you know what the `donated` column means?

server/src/controllers/contactController.ts
```diff
@@ -373,56 +342,61 @@ contactController.get('/orders/:id', async (req: Request, res: Response) => {
...
+    contact.orders.forEach((order) => {
+      const orderItemsMap = new Map<string, any>();
+      const {
+        ordertotal,
+        refunded,
+      } = order
+          .order_ticketitems
+          .reduce<{ordertotal: number, refunded: boolean}>((acc, ticket) => {
+            if (!ticket.ticketitem) return acc;
+            const key = `${ticket.ticketitem.ticketrestriction.eventinstanceid_fk}T${ticket.ticketitem.ticketrestriction.tickettypeid_fk}`;
```
henrykaus, line 357: You may also want to add the price to the key, there is a case where an admin may want to purchase two of the same ticket types for the same event, but set their price differently.

server/src/controllers/eventController.service.ts
```diff
@@ -40,67 +37,109 @@ export const createStripeCheckoutSession = async (
...
+      frequency: freq.one_time,
+    });
+    donationCartRows.push(
+        getCartRow('Donation', 'A generous donation', donation*100, 1),
+    );
+    donationTotal+=donation;
+  });
+
+  return {donations, donationCartRows, donationTotal};
+};
```
henrykaus, line 83: Is there a time where there could be more than one donation?

server/src/controllers/eventController.ts
```diff
@@ -56,70 +57,74 @@ export const eventController = Router();
...
-          donation,
-        donation ? cartRows.concat(donationItem) : cartRows,
-        orderTotal,
-        discount,
+          ticketCartRows.concat(donationCartRows),
+          {...discount, amountOff: discountAmount},
       );
+    } else if (ticketTotal + donationTotal > 0) {
+      return res.status(400).json({error: 'Cart Total must either be $0.00 USD or greater than $0.49 USD'});
     }
```
henrykaus, line 97: Do we need a case where the order total is negative?

server/prisma/migrations/20240131062714_update_order_process/migration.sql
```diff
@@ -0,0 +1,212 @@
...
+-- CreateTable
+CREATE TABLE "refunditems" (
+    "id" SERIAL NOT NULL,
+    "refundid_fk" INTEGER NOT NULL,
+    "order_ticketitemid_fk" INTEGER,
+    "donationid_fk" INTEGER,
+    "amount" MONEY NOT NULL,
+
+    CONSTRAINT "refunditems_pkey" PRIMARY KEY ("id")
+);
```
ben16williams, line 164: What's the difference between these? Why do they need to be distinct tables? Presumably there is the case where we would want to be able to partially refund an order, would that result in multiple `refund_intents` from stripe? Then we would want to keep track of each one and the associated amount

server/prisma/migrations/20240131062714_update_order_process/migration.sql
ben16williams: There is a workflow were if a customer knows that they won't be able to make it to the show, instead of having the ticket refunded they can 'donate' it so that we don't count it against the number of remaining tickets. Functionally that means you can 'oversell' a show but that the physical number of people who show up wont ever be more then the set total of available tickets

server/prisma/migrations/20240131062714_update_order_process/migration.sql
ben16williams: I agree, this seems like a good opportunity to start fresh with a new schema, ergo delete all the migrations and recreate the deployed DBs. We'd have to coordinate with others working on schema changes though

ben16williams: That would also make this a little better to review

ben16williams:
This is understandably quite a large PR, is there any way of modularizing it a bit? I realize such a fundamental schema change makes that difficult with out breaking things between PRs

client/src/components/Ticketing/ticketingmanager/DoorList/DoorList.tsx
benfaustin95: Fixed

server/prisma/migrations/20240131062714_update_order_process/migration.sql
benfaustin95: Fixed

server/prisma/migrations/20240131062714_update_order_process/migration.sql
benfaustin95: Updated

server/prisma/migrations/20240131062714_update_order_process/migration.sql
benfaustin95: Updated

server/prisma/migrations/20240131062714_update_order_process/migration.sql
benfaustin95: It was an existing events field. Currently it can not be changed, but once subscriptions are added in it will exclude an event from being included within the subscription.

server/prisma/migrations/20240131062714_update_order_process/migration.sql
benfaustin95: Fixed

server/src/controllers/contactController.ts
benfaustin95: Fixed

server/src/controllers/eventController.service.ts
benfaustin95: Updated.

server/prisma/migrations/20240131062714_update_order_process/migration.sql
benfaustin95: Currently the schema doesn't support multiple refunds per order, this update would allow for zero to many refunds per order with each refund having one to many refund items. A refund Item would contain the id of the item being refunded as well as the amount of  the original price that was refunded for that individual item (I was thinking this would allow for partial refunds on an individual item level if that ever was a need down the line).

server/src/controllers/eventController.ts
benfaustin95: No cart item itself can be negative so I think we might be fine.

henrykaus (dismissed):
All of my concerns have been addressed so this looks good to me. I think @ben16williams should probably give this one last look before merging due to its size.

server/prisma/migrations/20240210185042_add_refund_constraint/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- This is an empty migration.
+ALTER TABLE "refunditems" ADD CONSTRAINT "refund_ticketitems_check" CHECK (("donationid_fk" is null AND "order_ticketitemid_fk" is not null) OR ("order_ticketitemid_fk" is null AND "donationid_fk" is not null));
```
ben16williams, line 2: Does this need to be a separate migration? The comment should also be removed or altered

server/prisma/schema.prisma
```diff
@@ -128,61 +118,68 @@ model events {
...
-  discounts         discounts?   @relation(fields: [discountid_fk], references: [discountid], onDelete: NoAction, onUpdate: NoAction, map: "orders_discountid_fkey")
-  date              date         @relation(fields: [orderdate], references: [dateid], onDelete: NoAction, onUpdate: NoAction)
+  payment_intent    String?             @unique @db.VarChar(255)
+  checkout_sessions String?             @unique @db.VarChar(255)
+  contacts          contacts            @relation(fields: [contactid_fk], references: [contactid], onDelete: Restrict, onUpdate: Cascade, map: "orders_contactid_fkey")
+  discounts         discounts?          @relation(fields: [discountid_fk], references: [discountid], onDelete: Restrict, onUpdate: Cascade, map: "orders_discountid_fkey")
+  donation          donations?
+  refunds           refunds[]
+  order_ticketitems order_ticketitems[]
+}
```
ben16williams, line 164: I think we would still want to include the `ordertotal` in the order instead of having to fetch all the `orderitems` and check/apply a discount code to get the total.

server/prisma/migrations/20240210185024_init/migration.sql
```diff
@@ -1,3 +1,292 @@
...
+);
+
+-- CreateTable
+CREATE TABLE "order_ticketitems" (
+    "id" SERIAL NOT NULL,
+    "orderid_fk" INTEGER NOT NULL,
+    "price" MONEY NOT NULL,
+
+    CONSTRAINT "order_ticketitems_pkey" PRIMARY KEY ("id")
+);
```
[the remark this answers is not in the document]
ben16williams: This table appears to just be an intermediary between an order and a ticketitem, so could we skip it altogether and record the price with the ticketitem instead? And side note for a future PR, the idea of another table for recording ticket exchanges makes sense to me where we include a fk to it in the ticketitem as well

ben16williams (changes requested):
Left a few comments regarding schema changes. Over all I think this looks like a more manageable way of tracking tickets with room to expand the schema where needed

server/prisma/migrations/20240210185024_init/migration.sql
```diff
@@ -1,3 +1,292 @@
...
+    "id" SERIAL NOT NULL,
+    "message" TEXT NOT NULL,
+    "stack" TEXT NOT NULL,
+    "createdAt" INTEGER NOT NULL,
+
+    CONSTRAINT "Error_pkey" PRIMARY KEY ("id")
+);
+
 -- CreateTable
 CREATE TABLE "pphAccounts" (
```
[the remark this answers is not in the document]
benfaustin95: Updated, I also removed the seeding scripts but left the seeder data because I wasn't sure if that was needed.

server/prisma/migrations/20240210185042_add_refund_constraint/migration.sql
benfaustin95: Currently Prisma does not support constraints so the refund constraint needs to be added manually. I added it to the init migration, in the future we just need to remember to manually add it back if we ever get rid of the init migration.

server/prisma/schema.prisma
benfaustin95: I added an ordersubtotal and discounttotal.

server/prisma/migrations/20240210185024_init/migration.sql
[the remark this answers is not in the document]
benfaustin95: Currently I have both `orderticketitems` and `subscriptionticketitems` (will be added in another PR) "derived" from `ticketitems` because they share the same base data but differ in what they are related to., but that can be changed.

server/prisma/migrations/20240210185042_add_refund_constraint/migration.sql
ben16williams: Ah interesting that prisma does not support contraints. The idea with this is that we won't have to start over the migrations and once we go into production, we won't be able to

server/prisma/migrations/20240131062714_update_order_process/migration.sql
ben16williams: I see, I think we could still get away with flattening these, but since we already have a pattern of smaller tables mapped with lots of relations I think this is fine.

ben16williams (approved):
Looks like a good (re)start. Right before you merge this, you'll have to follow these instructions to delete the dev database so that the new migration gets applied fresh when dev is auto-deployed. 
[RESET DEV CLOUD SQL (1).pdf](https://github.com/WonderTix/WonderTix/files/14336216/RESET.DEV.CLOUD.SQL.1.pdf)

The same applies when we need to promote to staging and prod

Issue #555, opened 2024-01-06
Custom accommodation should appear in edit contact popup


henrykaus:
When a contact has a custom accommodation, the accommodation will not appear in the edit PopUp.

Expected behavior: Seating Accommodations should be set to `Other` and the Other Accommodation box should have the value that is stored with the contact.

<img width="847" alt="image" src="https://github.com/WonderTix/WonderTix/assets/92124260/a78c5eff-4ff2-430c-9a4c-0a1c7b461d3c">

PR #582, opened 2024-01-20
[WT-555] Fixed display/editing of custom seating accommodations in ContactPopUp

henrykaus:
### Description
You are now able to edit an accommodation that has been set from the Comments field.

image

### Validation
Manual testing.

### Issue
#555 

### Operating System
macOS M2


client/src/components/CRM/Contacts/ContactPopUp.tsx
```diff
@@ -34,6 +38,17 @@ const ContactPopUp = (props: ContactPopUpProps): ReactElement => {
     values,
   } = props;
 
+  // Assigns seatingAcc to comments input if there's a custom accommodation
+  const contact = {
```
benfaustin95, line 42: Is it possible to split the seating accommodation by the ' - ' so that the correct option can be selected upon load and the option itself is not in the comments edited by the user? Currently if the existing option is "Wheel Chair - ..."  and the user changes the option to Aisle the new seating accommodation will be "Aisle - Wheel Chair - ...".

henrykaus: Yeah, I think it probably makes the most sense to just create a `comments` column in the schema itself at this point, rather than keep reusing `seatingAccom` for not-related circumstances.

benfaustin95: That sounds good to me!

henrykaus: This PR grew quite a bit, but added the `comments` field

ben16williams: +1 for adding the comments field. I had concatenated the seating accommodation with the comments as a short term stop gap for recording any comments that users left in the checkout process since other wise they would have been ignored

benfaustin95 (approved):

PR #588, opened 2024-02-06
[WT-574] Fix refunded order items do not show in a contacts order history

ReidLuttrell:
### Description
When an order is refunded, it calls orderCancel, which sets refunded: true, but it also sets eventtickets to an empty array. This is what is causing the orderitems on the contact page to be empty after a refund.

### Risks
It could be that eventtickets is set to empty array for some other reason I'm unaware of, but if the intent is for the tickets bought to be shown in the contacts page after a refund, then this can't be done I believe.

### Validation
I made a test purchase, refunded that order, then verified that the refund was set to yes, and the order items were listed.

### Issue
[*Link to corresponding issue*](https://github.com/WonderTix/WonderTix/issues/574)

### Operating System
Windows 11

### DIFF
--- a/server/src/controllers/orderController.service.ts
+++ b/server/src/controllers/orderController.service.ts
@@ -171,18 +171,6 @@ export const orderCancel = async (
             },
             data: {
               refunded: true,
-              singletickets: {
-                update: item.singletickets.map((singleticket) => ({
-                  where: {
-                    singleticketid: singleticket.singleticketid,
-                  },
-                  data: {
-                    eventtickets: {
-                      set: [],
-                    },
-                  },
-                })),
-              },
             },
           }),
         ),


ben16williams:
I believe that the `eventtickets` are used to determine how many tickets are available for a show. If a ticket is refunded, then we don't want to count that as a `sold` ticket anymore. I didn't look at the logic but that does look to be the case after a cursory pulling down of the branch and refunding an order

ReidLuttrell:
> I believe that the `eventtickets` are used to determine how many tickets are available for a show. If a ticket is refunded, then we don't want to count that as a `sold` ticket anymore. I didn't look at the logic but that does look to be the case after a cursory pulling down of the branch and refunding an order

I've tried some approaches to fix this, but I'm pretty much at a loss as to how to fix this and preserve all the functionality. Mainly I tried to add a refunded field to the eventticket itself, so that I could say a ticket was sold if it had referenced eventtickets and it was not refunded, but `getOrderItems` will only consider a ticket available if its singleticket_fk is null I believe, and because of that, I see no obvious way to preserve the event data while also considering the ticket available, since the only way would be to have a singleticket_fk which leads to the event details, but that would make the ticket be considered as sold. To that effect I'm going to put this back to the backlog, and hopefully someone who is more familiar with the structure of the ticketing system can figure out a solution.

benfaustin95:
> > I believe that the `eventtickets` are used to determine how many tickets are available for a show. If a ticket is refunded, then we don't want to count that as a `sold` ticket anymore. I didn't look at the logic but that does look to be the case after a cursory pulling down of the branch and refunding an order
> 
> I've tried some approaches to fix this, but I'm pretty much at a loss as to how to fix this and preserve all the functionality. Mainly I tried to add a refunded field to the eventticket itself, so that I could say a ticket was sold if it had referenced eventtickets and it was not refunded, but `getOrderItems` will only consider a ticket available if its singleticket_fk is null I believe, and because of that, I see no obvious way to preserve the event data while also considering the ticket available, since the only way would be to have a singleticket_fk which leads to the event details, but that would make the ticket be considered as sold. To that effect I'm going to put this back to the backlog, and hopefully someone who is more familiar with the structure of the ticketing system can figure out a solution.

Issue #579, opened 2024-01-19
Save visit source with the contact


ben16williams:
The visit source (`how did you hear about us?`) field is not wired up at all in the checkout page. Need to create a new field for contacts and save the visit source during checkout (and admin checkout) if it's present

PR #592, opened 2024-02-12
[WT-579] Visit source contacts

Deacon-Coding:
### Description
added visitSource "How did you hear about us?" to contact.service and contact.router 

### Risks
none known

### Validation
changes appear in localhost

### Issue
https://github.com/WonderTix/WonderTix/issues/579

### Operating System
Windows 11


ben16williams:
I believe you'll need to add a `visitsource` string field to the `contacts` table for values to actually be saved. [Here's ](https://github.com/WonderTix/WonderTix/blob/main/server/prisma/schema.prisma#L11) the contact table in the schema that would updated and [here's ](https://github.com/WonderTix/WonderTix/pull/543#issuecomment-1873003064) a comment on how to do a DB migration with Prisma

server/src/api/contacts/contacts.service.ts
benfaustin95, line 1: We moved away from api v1 for contact information, I think this change might need to be made in `contactController.ts` and `eventController.service.ts` to have effect.

benfaustin95:
It looks like visitsource is not making its way into the database. In order to save visit source with the contact it needs to be added to the `validateContact` function `eventController.service.ts`.

server/src/controllers/eventController.service.ts
```diff
@@ -311,6 +311,7 @@ const validateContact = (formData: checkoutForm) => {
           new RegExp('^(\\+?\\d{1,2}\\s?)?\\(?\\d{3}\\)?[\\s.-]?\\d{3}[\\s.-]?\\d{4}$'),
       ),
     }),
+    ...(formData.visitSource && {visitSource: formData.visitSource}),
```
ben16williams, line 314: The new column name in `contacts` table is `visitsource` with a lowercase 's'. Updating this line to `...(formData.visitSource && {visitsource: formData.visitSource}),` should get things working

benfaustin95:
I think this just needs to be merged with the most recent updates to main and it should be good to go.

benfaustin95 (dismissed):

henrykaus (commented):
The v1 APIs were just removed as they aren't used anywhere. So all changes should occur in only the Prisma (v2) APIs.

henrykaus (approved):
Looks good!

Issue #576, opened 2024-01-18
Listen for Stripe refund events and update orders accordingly


ben16williams:
The webhook should also listen for and handle `charge.refund.updated` & `charge.refunded` events from Stripe. Wondertix should only mark an order has refunded when the one of these events is received from Stripe, not when the refund request is sent to Stripe from Wondertix.

PR #595, opened 2024-02-16
[WT-576] Listen to Webhook for Refund Updates

l-c-mcc:
### Description
Added new column to order and donations tables called "refund status". This tracks the status of the refund, and updates according to what is reported by the `charge.refunded` and `charge.refund.updated` webhooks.

Note that the database still immediately returns purchased seats to the purchasing pool, as I imagine that even if the refund fails, the customer would rather work to have the refund corrected than the seats back.

### Risks
* Right now, there is no way to handle failed refunds once reported to the database. I suspect such events would require manual intervention, so perhaps in the future when the task manager is fully set up, the webhook handler function could automatically create tasks when a refund fails (and cancel the tasks if a refund later succeeds).

### Validation
Manual testing

[These are some helpful card numbers for testing with.](https://docs.stripe.com/testing?testing-method=card-numbers#refunds)

### Issue
[WT-576](https://github.com/WonderTix/WonderTix/issues/576)

### Operating System
Pop OS 22.04


benfaustin95:
I am not 100% sure, but I think `stripe.refunds.create` only returns success if the refund has been completed successfully so we might not need to be listening via the web hook for refunds initiated from the ticketing manager.

l-c-mcc:
> I am not 100% sure, but I think `stripe.refunds.create` only returns success if the refund has been completed successfully so we might not need to be listening via the web hook for refunds initiated from the ticketing manager.
> https://docs.stripe.com/api/refunds/create

Just double checked, but there is a possibility (for example, using the card number `4000000000007726`) that the `stripe.refunds.create` will return pending. Success will only be reported through the webhook. There are also cases where success will initially be reported, and then later failure will come through the webhook. 

Also, just to be clear, can refunds be initiated from any where besides the ticketing manager? I'll need to make how the webhook handles refund events a bit more flexible if so.

l-c-mcc:
Updated branch to work with recent database changes

server/src/controllers/orderController.service.ts
```diff
@@ -39,6 +40,48 @@ export const ticketingWebhook = async (
...
+
+  await prisma.refunds.updateMany({
+    where: {
+      orderid_fk: order.orderid
+    },
+    data: {
+      refund_status: refundStatusState,
+    },
+  });
+};
```
benfaustin95, line 83: It looks like this function will update all refunds associated with a specific order to the same status, currently the db schema allows for multiple refunds per order so it might be worth updating the status based on the refund intent. I am not sure if the charge.refunded event sends the refund object, but it looks like the charge.refund.updated event sends the refund object so an easy way to go about it might be to grab the id (refund intent)  from that object and update the refund in our db based on that. Another possible route would be to grab the refund intent for all refunds associated with a payment intent then use `stripe.refunds.retrieve` to get their current statuses.

benfaustin95: Another thing we might want to do is delete all refund items associated with any failed refunds so that the status of associated order items is correctly represented in the system.

l-c-mcc: You are correct that charge.refunded does not send the refund object while the charge.refund.updated event does. And while across all my tests, I have always gotten both a charge.refunded event and a charge.refund.updated event for refunds, the [Stripe docs says that updated events only are sent for "selected payment methods", which I could not find a better description for.](https://docs.stripe.com/refunds#refund-webhook-events) As such, I implemented your second idea since that seemed more resilient.

server/src/controllers/orderController.service.ts
```diff
@@ -39,6 +42,66 @@ export const ticketingWebhook = async (
...
+      orderid_fk: order.orderid,
+    },
+    select: {
+      id: true,
+      refund_intent: true,
+      refund_status: true,
+    },
+  });
+
+  refunds.forEach(async (refund) => {
```
benfaustin95, line 79: I think we generally want to avoid using forEach with async calls it might be worth swapping this for a for..of loop.

server/src/controllers/orderController.service.ts
```diff
@@ -39,6 +42,66 @@ export const ticketingWebhook = async (
...
+    const refundObject = await stripe.refunds.retrieve(refund.refund_intent);
+    console.log(refundObject);
```
benfaustin95, line 81: Would you mind deleting the extra log?

server/src/controllers/orderController.service.ts
benfaustin95: Awesome thanks!

server/src/controllers/orderController.service.ts
l-c-mcc: I didn't know that, thanks! [Looks like map would allow for this part to be done in parallel, so I went with that.](https://stackoverflow.com/a/37576787)

server/src/controllers/orderController.service.ts
l-c-mcc: Oops, done

l-c-mcc:
Resolved comments, modified migrations to merge better with #619 given they change the same enum, and the refund status is now explicitly set to start at in_progress instead of that being the table default.

server/prisma/schema.prisma
```diff
@@ -218,6 +218,7 @@ model refunds {
   id            Int           @id @default(autoincrement())
   orderid_fk    Int
   refund_intent String        @unique @db.VarChar(255)
+  refund_status state         @default(in_progress)
```
benfaustin95, line 221: It looks like default value in the schema does not align with the migration.

l-c-mcc: in sync now

benfaustin95 (approved):

benfaustin95 (approved):

l-c-mcc:
Despite the two approvals, it seems I am unable to merge.

ben16williams:
> Despite the two approvals, it seems I am unable to merge.

Ben wasn't listed as a member of this repository. You should be able to merge now

Issue #596, opened 2024-02-17
UI Text on ticket selection too small


ReidLuttrell:
On the events page ticket selector, the UI text could be too small for elderly customers.

PR #597, opened 2024-02-18
[WT-596] Increased the size of the UI text for event showings and ticket selection

ReidLuttrell:
### Description
Increased the size of the UI text on the main ticket select because was told it could be too small for elderly customers

### Risks
Not sure how to test this on mobile but it should be enlarged there too

### Validation
Opened event showing view and inspected text

### Issue
https://github.com/WonderTix/WonderTix/issues/596

### Operating System
Windows 11

### DIFF
--- a/client/src/components/Ticketing/event/TicketPicker.tsx
+++ b/client/src/components/Ticketing/event/TicketPicker.tsx
@@ -277,7 +277,7 @@ const TicketPicker = (props: TicketPickerProps): ReactElement => {
   const promptMarkup = {
     selectDate: (
       <label
-        className='text-white font-semibold text-xl'
+        className='text-white font-semibold text-2xl'
         htmlFor='date-select'
       >
         Select date below ({props.tickets.length} showings)
@@ -289,7 +289,7 @@ const TicketPicker = (props: TicketPickerProps): ReactElement => {
         htmlFor='time-select'
       >
         {selectedDate ? format(selectedDate, 'eee, MMM dd') : ''}
-        <span className='text-white font-bold'> - Choose time:</span>
+        <span className='text-white font-bold text-2xl'> - Choose time:</span>
       </label>
     ),
     showSelection: (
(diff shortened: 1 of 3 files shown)


client/src/components/Ticketing/event/TicketPicker.tsx
```diff
@@ -462,7 +462,7 @@ const TicketPicker = (props: TicketPickerProps): ReactElement => {
           qty > selectedTicket.availableseats ||
           (selectedTicketType.name === 'Pay What You Can' && (payWhatPrice == null || payWhatPrice < 0))
         }
-        className='disabled:opacity-30 disabled:cursor-not-allowed py-2 px-3 mt-7 bg-blue-500 text-white enabled:hover:bg-blue-600 rounded-xl'
+        className='disabled:opacity-30 disabled:cursor-not-allowed py-2 px-3 mt-7 bg-blue-500 text-white enabled:hover:bg-blue-600 rounded-2xl'
```
henrykaus, line 465: Any particular reason why we're changing the rounding? It looks like most other buttons use `rounded-xl`

client/src/components/Ticketing/event/TicketPicker.tsx
```diff
@@ -289,7 +289,7 @@ const TicketPicker = (props: TicketPickerProps): ReactElement => {
         htmlFor='time-select'
       >
         {selectedDate ? format(selectedDate, 'eee, MMM dd') : ''}
-        <span className='text-white font-bold'> - Choose time:</span>
+        <span className='text-white font-bold text-2xl'> - Choose time:</span>
       </label>
```
henrykaus, line 293: For this section, it'd probably make more sense to apply the text sizing the the label, rather than just the ` - Choose time:` portion.

henrykaus (commented):
Just a couple smaller things

client/src/components/Ticketing/event/TicketPicker.tsx
ReidLuttrell: No, that was a mistake. Removed.

client/src/components/Ticketing/event/TicketPicker.tsx
ReidLuttrell: I added a commit where all the UI text is now consistently enlarged, including the option boxes and option labels. Let me know what you think. Thanks!

ReidLuttrell: Also thanks for the info about mobile sizing @henrykaus

henrykaus (approved):
Looks great!

Issue #447, opened 2023-11-13
Add the ability to upload a file for an event image


ben16williams:
PPH would like this as a convenient way to use custom artwork created by their marketing department. 

Once completed, the image url functionality can be removed.

ben16williams:
> Once completed, the image url functionality can be removed.

I take this back, we should keep both options

PR #604, opened 2024-02-25
[WT-447] Add ability to upload images on event creation and store in google cloud

ReidLuttrell:
### Description
This pull uses google cloud storage to add the ability to upload images to events on creation. On upload, the file is uploaded to the google cloud storage, where a link is generated and uses the existing URL input field and database column to keep track of the event images.

### Risks
Uploaded image links are public to anyone with the link

### Validation
Inserted custom image to create event and inspected events page

qxPIz8QTlg

### Issue
https://github.com/WonderTix/WonderTix/issues/447

### Operating System
Windows 11


server/src/controllers/eventController.ts
```diff
@@ -10,12 +10,21 @@ import {
...
 import {extendPrismaClient} from './PrismaClient/GetExtendedPrismaClient';
 import {isBooleanString} from 'class-validator';
 const prisma = extendPrismaClient();
 
+import multer from 'multer';
+import {Storage} from '@google-cloud/storage';
+
+const upload = multer();
+
+const storage = new Storage({keyFilename: './wondertix-app-12f15702b745.json'});
```
ben16williams, line 25: This should be an environment variable and we'll want to point to a different bucket for dev, staging, and prod. Ideally we don't have to use actually cloud storage for locally running environments. That would mean starting up a `fake` cloud storage image in docker that the locally running server talks to. But given the time frame, and given that we already use external services for local development (auth0), an acceptable solution would be to have locally running environments point to another shared bucket in cloud storage (eg. `image_upload_wondertix_local`).

ben16williams: Here's another PR that adds a new environment variable for reference https://github.com/WonderTix/WonderTix/pull/528

ReidLuttrell: The bucket name and bucket key file now use an env variable

README.md
```diff
@@ -38,21 +40,22 @@ Future features include managing/creating email campaigns and ticket exchanges.
...
-      - __NOTE__ - stripe listen _must_ remain running in your terminal for events to be forwarded and, subsequently, for the checkout process to complete. This means that everytime you wish to test the checkout process locally, you have to run the `stripe listen` command and leave it running. You should see your device listed in the Stripe developer portal on the `Wondertix Dev` account under `Webhooks -> Local listeners` and the `Status` should be `Listening`. 
-   4. You should now be able to sucessfully checkout with Stripe locally and see order details updated. See [Test Payment Methods](https://stripe.com/docs/testing) for how to checkout with Stripe in test mode. 
-6. The client will be available at <https://localhost:3000>
-7. The server will be available at <https://localhost:8000>
-8. The swagger docs will be available at <https://localhost:8000/api/docs>
+   3. Run `mkcert localhost` to create a certificate.
+4. Get the google cloud storage service account key file from your team lead and put into `<path/to/WonderTix/server>` for access to image upload.
```
ben16williams, line 48: Related to my earlier comment on `.env.dist`, configuring the service account key like this means that we would have no way to provide it to deployed environments without trying to do something fancier like mounting the file onto cloudrun which, like I mentioned, am not sure is possible and is not really necessary since this could just be another environment variable.

All environment variables used by deployed environments (dev,stg,prd) are stored in GCP's secret manager which hides the actual value. So, if the only way to get image uploads to work locally with an environment variable is to commit the secret value to `.env.dist`, then we could just not do that and image upload would only work for deployed environments once we add add the secret to secret manager.

If someone really wants to test out image upload locally, then they could do very similar to what is mentioned here already which would be to reach out to the TL and add the secret to there own `.env`.

Feel free to reach out if you have questions about what I'm trying to get at

ben16williams:
This functionally looks like it works pretty well! Can approve once my other comments are addressed

ben16williams (commented):
Just realized I had a pending review made from my phone the other day that I didn't submit 🤦

server/package.json
```diff
@@ -36,7 +36,9 @@
     "js-yaml": "^4.1.0",
     "json-refs": "^3.0.15",
     "morgan": "^1.10.0",
+    "multer": "^1.4.5-lts.1",
     "pg": "^8.11.2",
+    "react-dropzone": "^14.2.3",
```
henrykaus, line 41: Is `react-dropzone` necessary on the server side?

client/src/components/Ticketing/ticketingmanager/Event/components/EventGeneralForm.tsx
```diff
@@ -1,19 +1,55 @@
...
+    }
+  }, [setFieldValue]);
+
+  const {getRootProps, getInputProps} = useDropzone({onDrop});
+
+  return (
+    <div {...getRootProps()}>
+      <input {...getInputProps()}></input>
+      <p>Drag & Drop File or Click Here to Upload</p>
+    </div>
```
henrykaus, line 49: For this component, there should be some visual indicator (like a dotted line surrounding the drop zone) that shows that you can drag and drop here. Check out this tailwind-based component for possible styling: https://tailwindcomponents.com/component/dropzone-upload. You may want to make the component thinner than what's on this site though so it visually makes sense.

Let me know if you have any questions!

server/src/controllers/eventController.ts
```diff
@@ -143,6 +149,90 @@ eventController.post('/checkout', async (req: Request, res: Response) => {
...
+eventController.post('/image-upload', upload.single('file'), async (req: Request, res: Response) => {
+  const gcloudKey = process.env.GCLOUD_KEY;
+  const bucketName = process.env.GCLOUD_BUCKET;
+
+  if (gcloudKey === undefined || gcloudKey === '') {
+    console.log('Bucket key undefined!');
+    return res.status(500).send('Bucket key undefined!');
+  }
+  if (bucketName === undefined || bucketName === '') {
+    console.log('Bucket name undefined!');
```
henrykaus, line 190: I don't think the console.logs are necessary if the same message is being sent back in the response.

server/src/controllers/eventController.ts
```diff
@@ -143,6 +149,90 @@ eventController.post('/checkout', async (req: Request, res: Response) => {
...
+  }
+
+  try {
+    validateWithRegex(
+      req.file.mimetype,
+      'Invalid input, not a valid image filetype!', 
+      new RegExp('^(image\/(jpe?g|png))')
+    );
+  } catch (error) {
+    console.log(error);
```
henrykaus, line 210: For errors, we end up using `console.error` rather than `console.log`

server/src/controllers/eventController.ts
```diff
@@ -143,6 +149,90 @@ eventController.post('/checkout', async (req: Request, res: Response) => {
...
+  const file = imgBucket.file(req.file.originalname);
+  const stream = file.createWriteStream({
+    metadata: {
+      contentType: req.file.mimetype,
+    },
+    resumable: false,
+  });
+
+  stream.on('error', (err) => {
+    console.log(err);
```
henrykaus, line 223: Same here

server/src/controllers/eventController.ts
```diff
@@ -143,6 +149,90 @@ eventController.post('/checkout', async (req: Request, res: Response) => {
...
+
+  if (!req.file) {
+    return res.status(400).send('No file passed to request!');
+  }
+
+  try {
+    validateWithRegex(
+      req.file.mimetype,
+      'Invalid input, not a valid image filetype!', 
+      new RegExp('^(image\/(jpe?g|png))')
```
henrykaus, line 207: Does this regex also work for the capitalized versions of the extensions.

henrykaus (commented):
Had a larger comment on the look of the Drop Zone component, otherwise its looking good, with a few minor comments.

I noticed though that it doesn't look like any of these secrets have been put into the GCP Secret Manager yet, and so that would be required before this is deployed to any of the environments. I'm also not really able to fully locally test this on my own. Not completely sure where I'd get the `GCLOUD_KEY`.

server/package.json
ReidLuttrell: no, removed

client/src/components/Ticketing/ticketingmanager/Event/components/EventGeneralForm.tsx
ReidLuttrell: Dotted border has been added to indicate dropzone

server/src/controllers/eventController.ts
ReidLuttrell: removed these logs and made the logs in your other comments console.error

server/src/controllers/eventController.ts
ReidLuttrell: Filetype is now converted to lowercase before being regex'd

README.md
ben16williams: Talked with @ReidLuttrell and the proposed solution is to save the key file as a string in the gcp secret for deployed environments. Verified this works locally

ben16williams (approved):
Looks good to me. Nice job and thanks for seeing it through!

henrykaus (approved):
Thanks for all of the changes! It works great

PR #610, opened 2024-02-25
[WT-558-2] Add Subscription Type UI 

benfaustin95:
### Description

- Added subscription type UI for globally adding, editing, and removing subscription types. 
- Added preview slider to showings. 
- Added subscription eligibility slider to events. 
- Moved remaining Icon components to Icons.tsx (majority of file changes relate to changing Icon import location)
- Added subscription type seeder data

image
image

### Risks
Potentially an error in implementation logic 

### Validation
Manual testing of subscription types

### Issue
Second PR for issue 558

### Operating System
M1 Mac


client/src/components/Ticketing/Icons.tsx
```diff
@@ -215,6 +215,67 @@ const XIcon = (props: IconProps) => {
   );
 };
 
+const CirclePlusIcon = (props: {className?: string}) => {
```
henrykaus, line 218: Do you think you could structure the moved Icons similar to the others, to use the `IconProps`, have default values for `className` and `strokeWidth`, and then alphabetize them.

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypeUtils.tsx
```diff
@@ -0,0 +1,84 @@
...
+  };
+};
+
+export const getDefaultSubscriptionType = () => ({
+  id: -1,
+  description: '',
+  name: '',
+  price: 0,
+  previewonly: false,
+});
```
henrykaus, line 28: I think this could just be a constant object rather than a function call.
```
export const defaultSubscriptionType = {
  id: -1,
  description: '',
  name: '',
  price: 0,
  previewonly: false,
};
```

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
```diff
@@ -0,0 +1,374 @@
...
+          );
+        } catch (error) {
+          setPopUpProps(
+            'Delete Failed',
+            error.json
+              ? (await error.json()).error
+              : 'Error deleting subscription type',
+            false,
+            'delete-failed',
+          );
```
henrykaus, line 85: On an informational only popup, we usually only want to have the primary button, and leave out the secondary button and close button.

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
```diff
@@ -0,0 +1,374 @@
...
+      setPopUpProps(
+        'Save Failed',
+        error.message
+          ? error.message
+          : error.json
+          ? (await error.json()).error
+          : 'Subscription Type update failed',
+        false,
+        'save-failed',
+      );
```
henrykaus, line 167: Same here

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
```diff
@@ -0,0 +1,374 @@
...
+      },
+    },
+    {
+      field: 'actions',
+      headerName: '',
+      sortable: false,
+      width: 100,
+      renderCell: (cell) => {
+        return rowModesModel[cell.id]?.mode === GridRowModes.Edit ? (
+          <div className='w-[100%] flex flex-row justify-around'>
```
henrykaus, line 270: For consistency, do you think you could mirror these buttons after the same ones from the Ticket Types screen (styling, placement, and icons)

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
```diff
@@ -0,0 +1,374 @@
...
+      enterKeyDown,
+      tabKeyDown,
+    } = GridRowEditStopReasons;
+    if (
+      params.reason === rowFocusOut ||
+      params.reason === shiftTabKeyDown ||
+      params.reason === escapeKeyDown ||
+      params.reason === enterKeyDown ||
+      params.reason === tabKeyDown
+    ) {
```
henrykaus, line 108: is there any particular reason why these 4 button presses were specified? I believe it might end up hindering accessibility since you can't tab between fields and can no longer "enter" to save.

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypeUtils.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const validateSubscriptionType = (subscriptionType: any) => {
+  if (+subscriptionType.price <= 0) {
+    throw new Error('Invalid price');
+  }
+  if (subscriptionType.description.length === '') {
+    throw new Error('Invalid Description');
+  }
+  if (subscriptionType.name.length === '') {
+    throw new Error('Invalid Name');
+  }
```
henrykaus, line 13: For the errors that are displayed, I think it'd make sense for them to be a bit more descriptive. Like `Input Description` or `Please Input Description` or something around those lines. Also with the price, although it is an unlikely scenario, we probably don't want to have the arbitrary rule that they can't have a `$0` subscription price, so we'd just protect against negative prices.

client/src/components/Ticketing/Icons.tsx
benfaustin95: Fixed

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypeUtils.tsx
benfaustin95: Fixed

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
benfaustin95: Fixed

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
benfaustin95: Fixed

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
benfaustin95: Updated

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
benfaustin95: I think it might end up being confusing to have an explicit save button while also saving when the user leaves focus. It might be worth picking a singular option.

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypeUtils.tsx
benfaustin95: Updated

client/src/components/Ticketing/ticketingmanager/TicketTypes/SubscriptionTypes.tsx
henrykaus: I might be wrong, but I think if only rowFocusOut is left there then it’ll only save on button press or “enter” click

benfaustin95: I updated the handler so that enterKeyDown will save  👍

Issue #621, opened 2024-03-01
Add tasks to task manager when previously successful payments fail


l-c-mcc:
Most cases of previously successful payments later failing will likely need manual intervention. When such cases occur, adding them to the task manager would notify those who need to intervene.

Issue #620, opened 2024-03-01
Handle webhook events that notify issue with previous successful payment


l-c-mcc:
There are some Stripe events, delivered through the webhook, that carry notice that something went wrong with a previously successful payment, like a charge dispute. These different events need to be tracked down in the Stripe documentation and handled. [The test card page seems like a good place to start as some of those cards create such events](https://docs.stripe.com/testing).

Issue #619, opened 2024-03-01
Add order source and order status columns to orders table


l-c-mcc:
In PRs/issues being worked on, like the card readers (#587) and recurring donations (#600), the possibility of payments not being collected or taking a while to do so has increased. Thus, the status of orders should be tracked in the database. Tracking the source of the orders (admin checkout, card, reader, donation page, so on) will help more quickly determine the context of failed or stalled payments.

All completed transactions should be confirmed by webhook events.

PR #622, opened 2024-03-01
[WT-619] Add order source and order status columns to orders table

l-c-mcc:
### Description
Adds order_status and order_source columns to orders table and updates corresponding server code for reasons described in issue #619. Additionally, transitions to relying solely on order status column for order processing state. 

### Risks
Does not actually handle payments that have issues with them, but neither did the code before this, Arguably such cases are best dealt with in PRs relating to issues #620 and #621.
 
### Validation
Manual testing

### Issue
Issue #619 

### Operating System
Pop OS 22.04


server/src/controllers/orderController.ts
```diff
@@ -87,6 +87,7 @@ orderController.get('/refund', async (req: Request, res: Response) => {
     const orders = await prisma.orders.findMany({
       where: {
         payment_intent: {not: null},
+        order_status: state.completed,
         OR: [
```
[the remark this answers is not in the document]
l-c-mcc: Done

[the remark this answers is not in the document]
benfaustin95: Thanks!

server/src/controllers/orderController.service.ts
```diff
@@ -51,8 +54,25 @@ export const orderFulfillment = async (
...
+      break;
+    case 'card_reader':
+      orderSource = purchase_source.card_reader;
+      break;
+    case 'online_donation':
+      orderSource = purchase_source.online_donation;
+      break;
+    default:
+      break;
+  }
```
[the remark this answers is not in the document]
benfaustin95: Thanks!

benfaustin95:
It looks like this just needs to be updated with main and it should be good to go.

server/prisma/migrations/20240229214019_purchase_source/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- CreateEnum
+CREATE TYPE "purchase_source" AS ENUM ('online_ticketing', 'admin_ticketing', 'card_reader', 'online_donation');
+
+-- AlterEnum
+ALTER TYPE "state" ADD VALUE 'failed';
+
+-- AlterTable
+ALTER TABLE "orders" ADD COLUMN     "order_source" "purchase_source",
+ADD COLUMN     "order_status" "state" NOT NULL DEFAULT 'in_progress';
```
ben16williams, line 9: I would like the default value to be none/null if only to ever prevent a pattern of not explicitly setting the state on new orders

l-c-mcc: Done in latest commit.

l-c-mcc:
> It looks like this just needs to be updated with main and it should be good to go.

Now ready to merge into main. The biggest required here was in the readerWebhook function in orderController.service.ts.

benfaustin95:
This looks good to me, I am not sure if @ben16williams wants to take a last look.

ben16williams:
Just some merge conflicts that need addressed and then I'll approve

l-c-mcc:
Merge conflicts resolved

server/src/controllers/eventController.ts
```diff
@@ -888,35 +873,30 @@ eventController.use(checkScopes);
...
+    const response = await createStripePaymentIntent((ticketTotal + feeTotal) * 100);
+    return res.json(response);
   } catch (error) {
     console.error(error);
     if (error instanceof InvalidInputError) {
       res.status(error.code).json(error.message);
       return;
     }
-    res.status(500).json(error);
+    res.status(500).json('Internal Server Error');
```
henrykaus, line 899: Good catch, I've noticed we occasionally have an unsanitized error pop up on the client side.

server/src/controllers/orderController.service.ts
```diff
@@ -70,129 +74,151 @@ export const readerWebhook = async (
...
+        order_status: orderStatus,
+        order_source: orderSource,
+        checkout_sessions: checkoutSession,
         payment_intent: paymentIntent,
       },
     }),
     ...eventInstanceQueries,
   ]);
+
+  if (orderSubtotal - discountTotal === 0) {
```
henrykaus, line 201: I think you'd also need to add the `feeTotal` in this calculation.

server/src/controllers/orderController.ts
```diff
@@ -330,30 +331,32 @@ orderController.put('/refund/:id', async (req, res) => {
     if (!order) {
       return res.status(400).json({error: `Order ${orderID} does not exist`});
     }
-    if (!order.payment_intent) {
+    if (!order.payment_intent || order.order_status !== state.completed) {
```
henrykaus, line 334: Just curious but is there ever a case where there isn't a payment_intent but the state of an order could be completed?

server/src/controllers/orderController.ts
```diff
@@ -330,30 +331,32 @@ orderController.put('/refund/:id', async (req, res) => {
...
       return res.status(400).json({error: `Order ${orderID} has already been fully refunded`});
     }
 
     let refundIntent;
-    if (order.payment_intent.includes('comp')) refundIntent = `refund-comp-${order.orderid}`;
-    else if (order.payment_intent.includes('seeded-order')) refundIntent = `refund-seeded-order-${order.orderid}`;
-    else {
+    // eslint-disable-next-line camelcase
+    const {payment_intent} = order;
+    if (payment_intent.includes('comp') || payment_intent.includes('seeded-order')) {
```
henrykaus, line 344: For sake of not writing code specifically for our test data, does it make sense to just seed the orders with `comp` payment_intents, so they'd just be refunded like comped orders.

server/src/controllers/orderController.ts
benfaustin95: I don't think so, I moved the everything over so `order_status` is the only indicator of the processing state. I left the  payment intent check there because it can possibly be null based on the db schema, but not our logic, so TS gives a possibly null error.

server/src/controllers/orderController.service.ts
benfaustin95: Fixed, thanks for catching that!

henrykaus (approved):
Looks good! Thanks for fixing all of the indentation and simplifying the checkout process!

PR #625, opened 2024-03-08
added WEBSOCKET_URL var to build-client

ReidLuttrell:
### Description
Small change to add the react web socket env var as a build argument


ben16williams (approved):

l-c-mcc:
@ben16williams 

After this PR, we had to do one more step to get the websocket working: we had to update what events the Stripe webhook listened for on the dev account dashboard. If staging and production use different webhooks, they'll need to be updated accordingly. I have included an image below of the events currently added. I also took this as an opportunity to also add the events that will be needed for PR #595.

webhook events

PR #626, opened 2024-03-09
[WT-585] Added Discount Soft Deletion + UI Changes

### LINKED ISSUE #585: Add discount soft deletion
Adding discount soft deletion should allow admins to "delete" discounts without affecting the integrity of a discounts foreign key connections.

henrykaus:
### Description
Added Discount Soft Deletion so discount codes that are linked to orders can still be effectively deleted (added `deletedat` column to table).

UI Changes:
* Made headers and footers sticky in the Discount and Contact PopUps so that the actions are always visible.

https://github.com/WonderTix/WonderTix/assets/92124260/26888909-4caa-4636-867a-5e8a26482f4c

* Removed dark mode styles from Drop Zone component (missed this in the PR), and aligned elements here.

Before:
image

After:
image

* Added filler text (`—`) when there is no data present in the Ticket Type and Discount Code data grids.
image

* Fixed inconsistent Swagger docs groupings

### Risks
May have not updated a discount code route properly.

### Validation
Deleted used discount codes and checked database.

### Issue
closes #585 

### Operating System
macOS M2


benfaustin95:
It looks good, I think the only change that might need to be made is updating the discount query in the validateDiscount function.

henrykaus:
> It all looks good to me, I think the only change that might need to be made is updating the discount query in the validateDiscount function.

Fixed!

benfaustin95 (approved):

PR #628, opened 2024-03-11
[WT-589] Sold Out Indicators

### LINKED ISSUE #589: Sold out showings and events should be marked descriptively
When a particular showing or an entire event is sold out, it should be marked explicitly rather than just not appearing on the screen.

henrykaus:
### Description
Added three levels of `SOLD OUT` indicators, depending on where tickets are sold out.
* Shows SOLD OUT (and disables option) if a specific event instance is sold out
* Shows SOLD OUT if an entire date is sold out, but allows a user to click through to view the sold out times
* Shows SOLD OUT on home page and in the event information page if the entire event is sold out. Also, in this case...
  1. The ticket picker only displays dates and times
  2. the button on the Hero page is relabeled to just `See Information` and is restyled,
  3. and the event image is dimmed.

image
image
image
image

### Risks
Possibly want to do something else when an event is sold out.

### Validation
Some manual testing.

### Issue
closes #589 

### Operating System
macOS M2


ben16williams:
> Shows SOLD OUT on home page and in the event information page if the entire event is sold out. Also, in this case...
The ticket picker is not shown,
the button on the Hero page is relabeled to just See Information and is restyled,
and the event image is dimmed.

I think we'll probably want to still show the ticket picker in case users want to see specific dates for whatever reason. For instance if they want to ask for an override to get tickets for a sold out show. I think changing the button to 'show information' works well but I wouldn't be against renaming 'select date time' to 'view showings' or something too

henrykaus:
> > Shows SOLD OUT on home page and in the event information page if the entire event is sold out. Also, in this case...
> > The ticket picker is not shown,
> > the button on the Hero page is relabeled to just See Information and is restyled,
> > and the event image is dimmed.
> 
> I think we'll probably want to still show the ticket picker in case users want to see specific dates for whatever reason. For instance if they want to ask for an override to get tickets for a sold out show. I think changing the button to 'show information' works well but I wouldn't be against renaming 'select date time' to 'view showings' or something too

What do you think about this?
image

ben16williams:
> What do you think about this?

Could the sold out showings include the time? Otherwise looks good to me

ben16williams (approved):
looks good to me

PR #630, opened 2024-03-12
Put react-scripts back into dependencies

### LINKED ISSUE #536: Upgrade/fix vulnerable npm packages
# WonderTix Vulnerable NPM Packages

***

### Client npm audit:

```
@adobe/css-tools  <4.3.2
Severity: moderate
@adobe/css-tools Improper Input Validation and Inefficient Regular Expression Complexity - https://github.com/advisories/GHSA-prr3-c3m5-p7q2
fix available via `npm audit fix`
node_modules/@adobe/css-tools

axios  0.8.1 - 1.5.1
Severity: moderate
Axios Cross-Site Request Forgery Vulnerability - https://github.com/advisories/GHSA-wf5p-g6vw-rhxx
fix available via `npm audit fix --force`
Will install axios@1.6.2, which is a breaking change
node_modules/axios

nth-check  <2.0.1
Severity: high
Inefficient Regular Expression Complexity in nth-check - https://github.com/advisories/GHSA-rp65-9cf3-cjxr
fix available via `npm audit fix --force`
Will install react-scripts@3.0.1, which is a breaking change
node_modules/svgo/node_modules/nth-check
  css-select  <=3.1.0
  Depends on vulnerable versions of nth-check
  node_modules/svgo/node_modules/css-select
    svgo  1.0.0 - 1.3.2
    Depends on vulnerable versions of css-select
    node_modules/svgo
      @svgr/plugin-svgo  <=5.5.0
      Depends on vulnerable versions of svgo
      node_modules/@svgr/plugin-svgo
        @svgr/webpack  4.0.0 - 5.5.0
        Depends on vulnerable versions of @svgr/plugin-svgo
        node_modules/@svgr/webpack
          react-scripts  >=2.1.4
          Depends on vulnerable versions of @svgr/webpack
          Depends on vulnerable versions of resolve-url-loader
          node_modules/react-scripts

postcss  <8.4.31
Severity: moderate
PostCSS line return parsing error - https://github.com/advisories/GHSA-7fh5-64p2-3v2j
fix available via `npm audit fix --force`
Will install react-scripts@3.0.1, which is a breaking change
node_modules/resolve-url-loader/node_modules/postcss
  resolve-url-loader  0.0.1-experiment-postcss || 3.0.0-alpha.1 - 4.0.0
  Depends on vulnerable versions of postcss
  node_modules/resolve-url-loader

10 vulnerabilities (4 moderate, 6 high)
```

### Server npm audit:

```
axios  0.8.1 - 1.5.1
Severity: moderate
Axios Cross-Site Request Forgery Vulnerability - https://github.com/advisories/GHSA-wf5p-g6vw-rhxx
fix available via `npm audit fix`
node_modules/axios

1 moderate severity vulnerability
```

Research and refactoring will be needed, especially for the `client` project.

henrykaus:
_Note: this PR reverts a part of the changes from https://github.com/WonderTix/WonderTix/pull/617_

### Description
Due to the build process that we currently use, `react-scripts` must be labeled as a `dependency` within `client/package.json`. Currently the build process runs a `npm install --omit=dev` so that the build process can be quick. This means that `react-scripts` must be considered a normal dependency to run `npm build` when a production binary is created.

Fortunately, it looks like all of the "vulnerabilities" that `npm audit` displays for `react-scripts` are false alarms, and when `npm build` is run, it is doubtful that the package would even be included since it isn't used anywhere in the react code -- just in the build process.

In short, these vulnerabilities shouldn't be a problem, and we'll just have to ignore the "vulnerabilities" output stemming specifically from `react-scripts`. Output from other packages shouldn't be ignored though.

closes #536 again

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -39,6 +39,7 @@
         "react-redux": "^8.0.2",
         "react-router": "^6.2.1",
         "react-router-dom": "^6.2.1",
+        "react-scripts": "^5.0.1",
         "react-scroll": "^1.8.7",
         "react-use-websocket": "^3.0.0",
         "redux-mock-store": "^1.5.4",
@@ -61,7 +62,6 @@
         "eslint-plugin-jsdoc": "^39.6.2",
         "eslint-plugin-react": "^7.30.0",
         "prettier": "^3.0.3",
-        "react-scripts": "^5.0.1",
         "tailwindcss": "^3.1.4"
       },
       "engines": {
@@ -78,7 +78,6 @@
       "version": "2.2.0",
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
       "integrity": "sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==",
-      "devOptional": true,
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.1.0",
         "@jridgewell/trace-mapping": "^0.3.9"
@@ -129,7 +128,6 @@
       "version": "7.20.14",
       "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.20.14.tgz",
       "integrity": "sha512-0YpKHD6ImkWMEINCyDAD0HLLUH/lPCefG8ld9it8DJB2wnApraKuhgYTvTY1z7UFIfBTGy5LwncZ+5HWWGbhFw==",
-      "devOptional": true,
       "engines": {
         "node": ">=6.9.0"
       }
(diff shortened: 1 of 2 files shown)


ben16williams:
Why does the client need `react-scripts` to build in some cases but not others? If no part of the client is using that package then why can't it be removed outright or the deploy process adjusted so that it can be removed?

henrykaus:
> Why does the client need `react-scripts` to build in some cases but not others? If no part of the client is using that package then why can't it be removed outright or the deploy process adjusted so that it can be removed?

With our build process, `react-scripts` is always necessary to create a client build. `npm run build` is essentially just a shortcut for `react-scripts build`. It looks like the following in the `client/package.json`:

```
"build": "DISABLE_ESLINT_PLUGIN=true react-scripts build",
```

`react-scripts` is essentially a build tool, so it's not necessary once the client is built, but is always necessary to build the client (at least because we're using it as our build tool).

There is probably more than one way to build a react-app, but that would require a bit of a dive to swap to something different.

ben16williams:
> react-scripts is essentially a build tool, so it's not necessary once the client is built, but is always necessary to build the client (at least because we're using it as our build tool).

Ah makes sense👍

ben16williams (approved):

Issue #443, opened 2023-11-12
Ticketing - Resize image in Admin Events and Admin Season Page properly


henrykaus:
The image in the Admin Events Page and Admin Seasons Page should resize responsively to the size of the page. Currently it is taking up an odd amount of space, and does not respond to the current screen size.

![image](https://github.com/WonderTix/WonderTix/assets/92124260/89644431-1c2b-4972-919f-858050ada7af)

omariio1000:
Seems to me to be fixed. Had to just remove the max width. Responds to size of the window.

<img width="1176" alt="image" src="https://github.com/WonderTix/WonderTix/assets/43871473/c1273f8e-8a92-42aa-9bc5-1b0d882b6fe4">

PR #634, opened 2024-03-19
[WT-443] Ticketing - Resize image in Admin Events and Admin Season Page properly

omariio1000:
### Description
Image is resized according to screen size like described in #443.

### Risks
May want to set another max width or height.

### Validation
Manually tested

### Issue
Issue #443 

### Operating System
Windows 11 22H2 (23631.3296)

### Before:
image

### After:
image


henrykaus (commented):
It appears that the image is actually being sized too large in these cases. It looks like the image is going past the lowest hanging elements (the switches) in the information box.

omariio1000:
@henrykaus Added a max height, here is what it looks like now.
image
Not sure how to tweak it to match the height of the lowest hanging elements exactly, but can keep looking into it.

henrykaus (approved):

Issue #637, opened 2024-03-24
Only persist a new contact if the payment process is successful


ben16williams:
Create a new contact or modify an existing contact only when the Stripe Checkout session returns with a successful payment. 

In general, user level requests to the server should probably not be able to add or modifying any existing data. Ergo users should only have access to `GET` requests. This is intended as a stop-gap for https://github.com/orgs/WonderTix/projects/3/views/1?pane=issue&itemId=57488567

PR #648, opened 2024-04-05
[WT 637] Only persist contact if the payment process is successful 

benfaustin95:
### Description
- Updated  checkout so contact info is only persisted if order is successful 
- Updated contact controller so that all contact routes are protected 
- Converted newsletter field to timestamp

### Risks
Potentially missed updating a route use in the frontend . 

### Validation
Manual testing 

### Issue
#637 

### Operating System
M1 Mac


server/prisma/migrations/20240405001239_update_newsletter/migration.sql
```diff
@@ -0,0 +1,9 @@
+/*
+  Warnings:
+
+  - The `newsletter` column on the `contacts` table would be dropped and recreated. This will lead to data loss if there is data in the column.
+
+*/
```
henrykaus, line 6: I think you can manually remove this warning

server/src/controllers/eventController.service.ts
```diff
@@ -363,48 +361,80 @@ interface checkoutForm {
...
+  donorbadge?: boolean;
+  vip?: boolean;
+  volunteerlist?: boolean;
 }
 
 export const updateContact = async (
-    formData: checkoutForm,
     prisma: ExtendedPrismaClient,
+    contact: ValidatedContact,
+    check = 0,
```
henrykaus, line 372: Do you think you could rename the `check` parameter to be a bit more descriptive of its use. And/or instead of using integers, we could use strings that explain what is being checked

server/src/controllers/eventController.service.ts
```diff
@@ -363,48 +361,80 @@ interface checkoutForm {
   seatingAcc: string;
   comments: string;
   optIn: boolean;
+  donorbadge?: boolean;
+  vip?: boolean;
+  volunteerlist?: boolean;
```
henrykaus, line 366: It might be a good idea to renamed `checkoutForm` to something else since it's being used for multiple purposes besides just the checkoutForm. Probably would be a good idea to also switch from `optIn` to just `newsletter` at some point for consistency.

server/src/controllers/eventController.ts
```diff
@@ -79,7 +80,7 @@ eventController.post('/checkout', async (req: Request, res: Response) => {
       await validateDiscount(discount, ticketCartItems, prisma);
     }
 
-    const {contactid} = await updateContact(formData, prisma);
+    const updatedContact= validateContact(formData);
```
henrykaus, line 83: Since the contact isn't updated should this just be `validatedContact`?

server/src/controllers/eventController.ts
```diff
@@ -116,37 +117,37 @@ eventController.post('/checkout', async (req: Request, res: Response) => {
 
     if (orderSubTotal + feeTotal - discountAmount > .49) {
       toSend = await createStripeCheckoutSession(
-        contactid,
-        formData.email,
-        cartRows,
-        {...discount, amountOff: discountAmount},
+          updatedContact,
+          ticketCartRows.concat((donationCartRow? [donationCartRow]: []).concat(subscriptionCartRows)),
```
henrykaus, line 121: I think the `ticketCartRows.concat(...` is reverting a change from the last PR that merged. I think it is supposed to still be `cartRows`

henrykaus (commented):
One last thing, in the `ContactCard` could you change the `newsletter` field to saying something like `Since <date>` or `No` depending on the value of `newsletter`

henrykaus (approved):
Tested and looks good! Thanks for all of the changes!

PR #651, opened 2024-04-17
[WT-629] All discount codes are now Upper Case

### LINKED ISSUE #629: Make discount codes case-insensitive
Currently codes like `TENOFF` and `tenoff` have different meanings where they should instead be the same discount code. This may require changes to the schema and/or input validation.

alkindy:
### Description
Made discount codes creation/editing UPPER CASE, and retrieval case insensitive.

### Risks
Might be an area that discount codes are used that missed this treatment.

### Validation
Created, edited, and entered discount codes to check if they automatically switch to upper case.

### Issue
Closes #629 

### Operating System
Mac: Intel & M2


henrykaus (commented):
I think what is here so far looks good but it might be missing one major:
1) The backend routes in `discountController.ts` should be updated so that the `code` is brought `toUpperCase()` before it is used. This would include creation as well as just using.

henrykaus (commented):
I think the commit you reverted was on the right track. Most of the logic for the uppercase-ness should be server-side since client-side logic can be easily spoofed by directly accessing an endpoint.

henrykaus (commented):
@ben16williams @benfaustin95 This PR is now ready for review

benfaustin95 (approved):
Looks good!

PR #652, opened 2024-04-18
[WT-78] Extracting relevant data from "contacts" table for sending emails

lkhellah:
### Description
Extracted firstname, lastname, and email of users who have opted into emails


omariio1000:
PR no longer needed.

Issue #665, opened 2024-05-06
Update ticketing state to maintain accurate available seat count


benfaustin95:
Current Issue: Right now, customers can add more tickets to their cart than there are available seats for a showing by selecting different ticket types. For instance, if there are 5 available seats for a showing, and there are 5 General Admission tickets and 5 Pay What You Can tickets available, a customer could add up to 5 of each type to their cart, exceeding the available seats.
Desired Behavior: Ideally we want to prevent customers from adding more tickets to their cart than there are available seats for a showing. This means the ticketing state should maintain and accurate count of available seats for each showing and prevent customers from selecting more tickets than the available quantity, regardless of the ticket type.

PR #667, opened 2024-05-06
[WT-665] Update ticketing state to maintain accurate available seat count

benfaustin95:
### Description
Update ticketing state to prevent customers from adding more tickets to their cart than there are available seats for a showing by selecting different ticket types.

Note: When merged with the PR for issue 640 if a customer has added the maximum number of tickets available for a showing to their cart the showing wont show up as sold out, but rather the customer will just not be able to add additional tickets.

### Risks
Implementation logic could be incorrect. 

### Validation
Manual testing

### Issue
#665 

### Operating System
M1 Mac


henrykaus (approved):
Manually tested and works!

Issue #548, opened 2024-01-02
Enhance contacts search functionality


ben16williams:
Add search boxes for filtering on email, phone number, ID, address, city, state, country, and postal code and split the name search box into one for first name and one for last name.

Have results appear in a similar format to the doorlist, with all the same contact information (plus newsletter signups) and the ability to export the results. Users should be able to select a contact so that they are redirected to the contacts page with order history. Stretch goal would be to be able to select which columns appear in the search result

Need to be able to filter on newsletter signups by date so that an admin can get all contacts who signed up for the newsletter after a certain date. This may mean adding a timestamp to contacts for the newsletter signup time

PR #668, opened 2024-05-06
[WT-548] Enhance contacts search functionality

benfaustin95:
### Description
Primary Changes:

- Enhance contact search functionality to support searching by multiple parameters.
- Revamp result presentation by utilizing data grid 

### Risks
It is possible I missed a search edge case, or that the page's functionality is not what was desired. 

### Validation
Manual testing

### Issue
#548 

### Operating System
M1 Mac


ben16williams:
This looks so much better but one blocking issue is that the `Rows per page` looks like it's no-op. It would be really nice to be able to view more then 10 contacts at a time 

Non-blocking issues are that navigating back to search from viewing the contact page clears search parameters and that searches have to match the value exactly instead of giving results that partially match

benfaustin95:
> This looks so much better but one blocking issue is that the `Rows per page` looks like it's no-op. It would be really nice to be able to view more then 10 contacts at a time
> 
> Non-blocking issues are that navigating back to search from viewing the contact page clears search parameters and that searches have to match the value exactly instead of giving results that partially match

I pushed some updates that should fix the "Rows per page" issue, as well address the non-blocking issues.

ben16williams:
lgtm can approve once the merge conflict is resolved

benfaustin95:
> lgtm can approve once the merge conflict is resolved

The merge conflicts should be resolved now 👍

ben16williams (approved):

PR #669, opened 2024-05-07
Make order and refund state database fields optional

ben16williams:
### Description
This should hopefully fix an error applying latest database migration in dev ([logs](https://console.cloud.google.com/logs/query;cursorTimestamp=2024-05-01T17:55:08.388472Z;endTime=2024-05-02T05:44:00.000Z;pinnedLogId=2024-05-06T23:24:29.184246Z%2F663966ad0002cfb6167f0bee;query=resource.type%20%3D%20%22cloud_run_revision%22%0Aresource.labels.service_name%20%3D%20%22wtix-server-dev%22%0Aresource.labels.location%20%3D%20%22us-west1%22%0A%20severity%3E%3DDEFAULT%0Atimestamp%3D%222024-05-01T17:55:00.632914Z%22%0AinsertId%3D%22663281f40009a852c065f35c%22;startTime=2024-05-01T13:00:00.000Z?project=wondertix-app))

### Risks
If there is logic that expects the order or refund state to not be null and isn't validating that it isn't first, this could cause some NPEs or otherwise blow something up 

### Validation
local deployment with seeded data comes up healthy with all migrations successfully applied

### Operating System
win10

### DIFF
--- a/server/prisma/migrations/20240507011154_make_status_fields_optional/migration.sql
+++ b/server/prisma/migrations/20240507011154_make_status_fields_optional/migration.sql
@@ -1,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "orders" ALTER COLUMN "order_status" DROP NOT NULL;
+
+-- AlterTable
+ALTER TABLE "refunds" ALTER COLUMN "refund_status" DROP NOT NULL;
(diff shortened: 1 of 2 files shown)


benfaustin95:
If keeping the fields as non-optional is something we want to do, we could try dropping the database applying the migrations and reseeding.

ben16williams:
> If keeping the fields as non-optional is something we want to do, we could try dropping the database applying the migrations and reseeding.

I assumed the issue was that the seeding data was trying to insert some data that was missing these newly required fields but I don't get any issues locally so I'll try bouncing the DB

henrykaus (commented):
Is this PR still necessary?

Issue #659, opened 2024-04-30
Remove references to Portland Playhouse


ben16williams:
With the multi-tenancy design for supporting different companies, Wondertix is more of it's own application that is separate from PPH. 

Generify the webpage by removing any reference to PPH (the back drop on the main page also needs to be removed/replaced)

Generify the codebase from any references to PPH or write follow up stories to do so.

baylie-r-g:
Would we need to create a WonderTix logo to replace the PPH one?

ben16williams:
Shouldn't be necessary but would be cool if you're feeling creative!

baylie-r-g:
Literally started working on one while waiting for an answer

baylie-r-g:
![WTix logo](https://github.com/WonderTix/WonderTix/assets/126998719/e9c59c82-8bc6-4802-8cd9-514c6ea2ddb4)
Not exactly professional quality but it should make a solid placeholder

baylie-r-g:
Do I need to get rid of the Copyright statement in the footer, or does it need to stay?

ben16williams:
Let's keep the copyright statement for now

PR #674, opened 2024-05-14
Ticket659: Replaced PPH-specific assets

baylie-r-g:
### Description
Created a new WTix logo to replace PPH logos. Link in admin navbar leads back to home
I found a more general Creative Commons photo to place on the banner, no licensing or credit is required for commercial or personal use. (no copyright issues)

Replaced as many PPH logos as I could in the ticketing manager. Unable to find line holding the image in the tab.

### Risks
Unable to find the line for the final logo in the Ticket Management tree

### Validation
Ran all main pages and admin pages of the WTix site. I was able to check the loading page as well.

### Issue
(https://github.com/WonderTix/WonderTix/issues/659)

### Operating System
Windows 11


client/tailwind.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
 
     extend: {
       backgroundImage: {
-        'hero': 'url(../src/assets/portland_playhouse_cover.jpg)',
+        'hero': 'url(../src/assets/pexels-cottonbro-6896179.jpg)',
```
ben16williams, line 21: I was thinking that we could make this even more generic. So something like a solid background color or a simple pattern that's on theme. Ideally Wondertix will be general enough that it wont be restricted to servicing playhouses. I also don't know there are copy right issues with this image?

client/public/index.html
```diff
@@ -8,8 +8,8 @@
       name="description"
       content="Web site created using create-react-app"
     />
-    <link rel="icon" href="2011_Logo_white.png" />
-    <link rel="apple-touch-icon" href="2011_Logo_white.png" />
+    <link rel="icon" href="WTix logo light.png" />
+    <link rel="apple-touch-icon" href="WTix logo light.png" />
```
ben16williams, line 12: Probably a non-issue but not having any spaces in the image file name would be on the safer side

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "react-swipeable-views": "^0.14.0"
+  }
+}
```
ben16williams, line 5: Is this really needed at the root level?

ben16williams:
Would you mind adding a little more descriptive title? It helps when looking back through old PRs

ben16williams:
I see that there are still a few reference to Portland Playhouse in the code-base that we'll want to remove. You should be able to do a search and find with most text editors (including vscode) to find the rest

client/tailwind.config.js
baylie-r-g: The image I placed there is Creative Commons so it is free to use in any way. I can switch to something more generic though

package.json
baylie-r-g: I didn't add this myself so I do not know if it is needed at root level or not

baylie-r-g:
New commits for this PR

client/public/index.html
baylie-r-g, line 1: Should I add %PUBLIC_URL% back into this file?

package.json
henrykaus: @benfaustin95 Do you know why this is showing up now? I believe you were using the swipeable-views at one point

benfaustin95: The package is still being used at the client level, but shouldn't be at the root level. It looks like the package.json is being added with this PR is it possible the package was installed at the wrong level which generated the the file?

henrykaus: Ok, in that case, the file should probably just be deleted from the root directory.

client/public/index.html
henrykaus: Yeah, I think that would be best

package-lock.json
henrykaus, line 1: This file should also be deleted since it shouldn't be around at the root level

baylie-r-g:
New push created. Need to know if there is a server to upload a new logo so that it can be changed on the auth0 login

client/src/components/Ticketing/ticketingmanager/udash_navbar.tsx
henrykaus, line 1: This file was removed in a recent PR, so you can delete this file.

client/src/components/CRM/Navigation.tsx
henrykaus, line 1: This file can also be removed

henrykaus (commented):
Two minor changes and then this looks good to go. I really like the new WonderTix logo!!

henrykaus (approved):
Thanks for all the changes! Feel free to merge once tests pass.

Issue #662, opened 2024-04-30
Redesign the main page


ben16williams:
This is entirely an exercise in UX design and React. The issue of accessibility in the main page has been brought up before, particularly in the ticket-picker where the text may be to small for some groups to be able to read is easily. The key issues with the main page to address:
1. The Event scroll bar should be replaced with a grid where all the active Events fit on the screen horizontally and the scrolling is vertical. 
2. The Event widgets need to be bigger so that the text can be larger and easier to read. 
3. After selecting an Event, the dropdown interface for selecting a showing is odd. A cleaner UI would be to have each showing date be it's own widget, then selecting a showing would bring up the dropdowns to select the ticket type and quantity. PPH ticketing website is a good example. 

The goal is make something that would feel familiar to anyone and that is intuitive to use

PR #677, opened 2024-05-19
[WT-662] Redesigned ticket picker interface (Part 1)

henrykaus:
### Description
_Note: This PR strictly covers the ticket picker interface (home page in a future PR)._

Redesigned the ticket picker interface with (ideally) the same functionality as before but with a more more intuitive (and appealing) interface.

Let me know if you have any suggestions, I know this differs a bit from the original concept.

image
image
image

 
image

### Risks
Might need extra testing around ticket limits to ensure not too many tickets can be added to the cart.

### Validation
Manual testing with a variety of scenarios.

### Issue
#662 

### Operating System
macOS M2


ben16williams:
This looks great, the only nit I have is that buttons to increase or decrease the ticket quantity should probably be swapped so that the increase button is on the right. That would match the orientation of those buttons in the cart as well

henrykaus:
> This looks great, the only nit I have is that buttons to increase or decrease the ticket quantity should probably be swapped so that the increase button is on the right. That would match the orientation of those buttons in the cart as well

Good catch, swapped them.

ben16williams (approved):

PR #684, opened 2024-05-31
[WT-662] Redesigned Main Page

### LINKED ISSUE #662: Redesign the main page

henrykaus:
### Description

**Changes:**
* Redesigned home page
* Added discount code search on enter-key press in cart
* Small adjustments to styling of ticket picker

**Home Page Features:**
* Max of two events per row, if there is 1 event on a row it will be left-aligned
* Displays the first and last date of an event (if all showings occur on one date, it will only show the one date)
* Has a Coming Soon section (events in this section are placed here if the start date is after the current date [determined client-side])
* Has a Now Playing section (events in this section are placed here if they aren't placed in the coming soon column)
  * If a section is empty, it will not be displayed
  * If both sections are empty, a message will be displayed to come back later
* Event descriptions are truncated at 215 characters, but are still fully available if the user drills into the event

image
image
image

### Risks
Shouldn't be much, but I did have to modify the event/slice API to calculate startdate, enddate, sort events by startdate, and not include events if they have no showings. 

### Validation
Manual testing.

### Issue
closes #662 

### Operating System
macOS M2


ben16williams (approved):
lgtm thanks for doing this, looks really good imo!

PR #685, opened 2024-06-04
[WT-678] Disallow Customer To Purchase Tickets for Past Showings

### LINKED ISSUE #678: Do not allow purchase of tickets for showings that have already occurred for customers
Customers should not be able to purchase tickets if the showing has already happened.

This should be implemented purely on the server-side in the following places: checkout and the `events/slice` endpoint (possibly also `event-instance/tickets` and `/ticket-restriction`).
* In checkout, the purchase should be invalidated if a purchased item took place at or after the start time for its corresponding showing (for customer-purchase only).
* In the `event-instance/tickets` endpoint, showings returned should be filtered out to not include any with start times before the current time.
* In the `events/slice` endpoint, events returned should be filtered out to not include any with the end date/time before the current time

henrykaus:
### Description
Modified the following areas to restrict customers from purchasing tickets for shows that have started (determined by comparing the server's current date and time to an event instance's start date/time. If the current time is the same as or after the start date/time, the showing should not be purchasable _by a customer_):
* `/ticket-restriction`, `/events/slice`, `/event-instance/tickets`, and `/events/checkout` check for whether the showing has passed
* Shifted order source from being sent from client-side to an endpoint to being determined by which API was hit (because you could easily spoof this value otherwise):
  * Created identical checkout endpoint `events/admin-checkout` to `events/checkout` that requires an admin token and sets the `order_source` to `admin_ticketing`
  * Modified checkout process to reject an order if the order_source was from `online_ticketing` and the showing already happened
* Created similar `/ticket-restriction/all` endpoint used by admin purchase to `/ticket-restriction` that returns all ticket restrictions regardless of whether the showing had passed, but requires admin token

### Risks
This could use more manual testing to verify that not too many or too little showings or ticket restrictions are available.

If the theatre and server are in different timezones, this may cause showings to go off sale too soon or too late.

### Validation
Manual testing of checkout, admin-checkout, and home screen.

### Issue
closes #678 

### Operating System
macOS M2


benfaustin95 (approved):
Looks good!

PR #687, opened 2024-06-12
[WT-541] Added Search by Name to Refunds Page

### LINKED ISSUE #541: Add ability to search by contact name in addition to contact email in the refunds page

henrykaus:
### Description
Added the ability to search by name to the Refunds page. It works like below:
1. If there are no spaces in-between words in the search query, the API returns all results where the email, first name, or last name contain the query
2. If there are spaces in the search query, the API returns all results where the first name contains the first word and the last name contains the second word (any other words are ignored)

**Other changes:**
* Orders in refunds page are sorted from newest to oldest
* Made minor improvements to the refunds interface to make the search bar more consistent with other element, and include the emails in the displayed data (now utilizing native HTML `table` to auto-size the columns).
* Entering empty string in search bar clears results rather than displayed error msg
* Made logo on main page a button to navigate to home

image

### Risks
N/A

### Validation
Manual Testing

### Issue
closes #541 

### Operating System
macOS M2

### DIFF
--- a/client/src/components/Ticketing/Icons.tsx
+++ b/client/src/components/Ticketing/Icons.tsx
@@ -507,6 +507,27 @@ const TrashCanIcon = (props: IconProps) => {
   );
 };
 
+const UserIcon = (props: IconProps) => {
+  const {className = 'h-5 w-5', strokeWidth = 2} = props;
+
+  return (
+    <svg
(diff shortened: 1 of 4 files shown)


benfaustin95 (approved):
Looks good!

